PR #2, opened 2022-12-08
Read by ID for Library and Measure services

elsaperelli:
# Summary
Implements read by ID for Library and Measure services.
By: Nat Jones and Elsa Perelli

## New Behavior
Queries the mongo database for the Library or Measure resource by the matching ID and exposes an endpoint for Library and Measure services.

## Code Changes
- `.env.example` - changed database url variable name to match the others.
- `.env.test` - added a separate env.test file.
- `package-lock.json/package.json` - added supertest as a dependency for unit testing.
- `envConfig.ts` - properly routes which .env file to use depending on whether you are running unit tests or not.
- `serverConfig.ts` - added profile for Library resource route.
- `Connection.ts` - added envConfig.
- `DBOperations.ts` - added `findResourceByID` function that queries the mongo database for the resource with the matching ID.
- `index.ts` - removed dotenv because it is now called in `Connection.ts`.
- `LibraryService.ts` - service implemented for the Library resource. Includes a `searchById` function.
- `MeasureService.ts` - added functionality to the `searchByID` function.
- `node-fhir-server-core.d.ts` - added the type of app.
- `errorUtils.ts` - added error messages for the unit tests (taken from the test server).
- `LibraryService.test.ts` - tests searchByID for Library service.
- `MeasureService.test.ts` - tests searchByID for Measure service.
- `index.test.ts` - RIP noop test.
- `utils.ts` - adds `createTestResource`, `cleanUpDb`, and `testSetup` functions for the unit tests.

# Testing Guidance
- `npm run test` - ensure all the unit tests pass.
- `npm run lint:fix` - make sure there are no lint errors.
- This is where it gets tricky because we have not yet implemented createResource...
- In `index.ts`, add the following lines of code after the first `.then` (line 18):
```
  .then(() => {
    Connection.db.collection('Measure').insertOne({ resourceType: 'Measure', id: 'measure123', status: 'active' });
    Connection.db.collection('Library').insertOne({ resourceType: 'Library', id: 'library123', status: 'active' });
  })
```
- `npm run db:reset`
- `npm start`
- In Insomnia, send a `GET` request to `http://localhost:3000/4_0_1/Measure/measure123` with the header Content-Type: application/json+fhir. Make sure you get a `200 OK`.
- In Insomnia, send a `GET` request to `http://localhost:3000/4_0_1/Library/library123` with the header Content-Type: application/json+fhir. Make sure you get a `200 OK`.
- Also, send a `GET` request to an endpoint we know doesn't exist (such as `http://localhost:3000/4_0_1/Library/InvalidID`) and make sure you get a `404 Not Found` error.
- Let Nat or I (Elsa) know if you have any questions about testing!


src/db/DBOperations.ts
```diff
@@ -0,0 +1,10 @@
+import { FhirResourceType } from '@projecttacoma/node-fhir-server-core';
+import { Connection } from './Connection';
+
+/**
+ * searches the database for the desired resource and returns the data
+ */
+export const findResourceById = async (id: string, resourceType: FhirResourceType) => {
+  const collection = Connection.db.collection(resourceType);
+  return collection.findOne({ id: id }, { projection: { _id: 0 } }) as any;
+};
```
mgramigna, line 10: For now I think we can just use the generic on the invocation of `findOne` to get type safety instead. Also, let's change this to an `export async function` instead of export `const`:


``` ts
export async function findResourceById<T extends fhir4.FhirResource>(id: string, resourceType: FhirResourceType) {
  const collection = Connection.db.collection(resourceType);
  return collection.findOne<T>({ id: id }, { projection: { _id: 0 } });
}
```

Then the services will get typesafety by passing in the resource interface as the generic:

``` ts
 const result = await findResourceById<fhir4.Library>(args.id, 'Library');
``` 

`result` will be of type `fhir4.Library | null`.

This is a fine compromise for now. Maybe we'll figure out a better solution later

src/config/serverConfig.ts
```diff
@@ -1,12 +1,16 @@
 import { constants, ServerConfig } from '@projecttacoma/node-fhir-server-core';
 import { MeasureService } from '../services';
+import { LibraryService } from '../services/LibraryService';
```
mgramigna, line 3: Suggestion: just add `export * from './LibraryService'` to `services/index.ts`, and that will allow you to combine these imports into the one on line 2

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
+import { initialize, Server } from '@projecttacoma/node-fhir-server-core';
+import { serverConfig } from '../src/config/serverConfig';
+import { cleanUpDb, testSetup } from './utils';
+const supertest = require('supertest');
```
mgramigna, line 4: Shouldn't need to use require here. If `import` gives you an error that says it can't find a declaration file, install they DefinitelyTyped types and it should be fine:


```
npm install --save-dev @types/supertest
```

The process should always be:

1. Ideally we don't get this error
2. If we do, hopefully there's a `@types/*` repo for it, try to install it
3. If there isn't one, hopefully it's simple enough to write our own `d.ts` file for it
4. If 1-3 fail, use require

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    await testSetup([Library]);
+  });
+  describe('CRUD operations', () => {
+    test('test searchById with correctHeaders and the id should be in database returns 200', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library/library123')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then((response: any) => {
```
mgramigna, line 28: When you install the supertest types, methinks this will no longer need the `any` type declaration

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+          expect(response.body.id).toEqual(Library.id);
+        });
+    });
+    test('test searchById when the id cannot be found in the database', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library/invalidID')
+        .set('Accept', 'application/json+fhir')
+        .expect(404)
+        .then((response: any) => {
```
mgramigna, line 37: When you install the supertest types, methinks this will no longer need the `any` type declaration

test/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const dbUrl = process.env.DATABASE_URL || 'mongodb://localhost:27017/measure-repository';
+
+const createTestResource = async (data: any, resourceType: string) => {
+  const collection = Connection.db.collection(resourceType);
+  await collection.insertOne(data);
+  return { id: data.id };
+};
+
+//clean up db after test
```
mgramigna, line 11:
```suggestion
```

Function is intuitive enough

test/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { Connection } from '../src/db/Connection';
+
+const dbUrl = process.env.DATABASE_URL || 'mongodb://localhost:27017/measure-repository';
```
mgramigna, line 3: Should this default to the same one in `.env.test`? We don't want to accidentally use the real database during tests

```suggestion
const dbUrl = process.env.DATABASE_URL || 'mongodb://localhost:27017/measure-repository-test';
```

test/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { Connection } from '../src/db/Connection';
```
mgramigna, line 1: Can we convert all the export consts to export functions for consistency?

test/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const dbUrl = process.env.DATABASE_URL || 'mongodb://localhost:27017/measure-repository';
+
+const createTestResource = async (data: any, resourceType: string) => {
+  const collection = Connection.db.collection(resourceType);
+  await collection.insertOne(data);
+  return { id: data.id };
+};
```
mgramigna, line 9: Does this work?

```suggestion
const createTestResource = async (data: fhir4.FhirResource, resourceType: FhirResourceType) => {
  const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
  await collection.insertOne(data);
  return { id: data.id };
};
```

Remember to not use `any` unless you really have to. The above seems to work for me without running into those crazy issues we got yesterday (that must've been unique to `.find`

test/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+//clean up db after test
+export async function cleanUpDb() {
+  await Connection.db.dropDatabase();
+  await Connection.connection?.close();
+}
+
+export const testSetup = async (testfixtureList: any[]) => {
```
mgramigna, line 17:
```suggestion
export const testSetup = async (testFixtureList: fhir4.FhirResource[]) => {
```

test/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  await Connection.connect(dbUrl);
+
+  const result = testfixtureList.map(async x => {
```
mgramigna, line 20:
```suggestion
  const result = testFixtureList.map(async x => {
```

src/index.ts
```diff
@@ -1,10 +1,7 @@
 import { initialize, loggers } from '@projecttacoma/node-fhir-server-core';
-import * as dotenv from 'dotenv';
 import { serverConfig } from './config/serverConfig';
 import { Connection } from './db/Connection';
 
-dotenv.config();
```
mgramigna, line 6: why did this get yoinked from index.ts and into the db connection? I feel like we want to call dotenv as far up as possible (i.e. the entrypoint of the server).

src/services/LibraryService.ts
```diff
@@ -0,0 +1,29 @@
+import { RequestArgs } from '@projecttacoma/node-fhir-server-core';
+import { findResourceById } from '../db/DBOperations';
+import { Service } from '../types/service';
+import { ResourceNotFoundError } from '../util/errorUtils';
+
+/*
+ * Example service that one would implement for the `Library` resource
+ * See the Service interface for all possible functions
+ */
```
mgramigna, line 9: Guess this is no longer an example now that we have something huh?

src/services/MeasureService.ts
```diff
@@ -1,4 +1,7 @@
+import { RequestArgs } from '@projecttacoma/node-fhir-server-core';
+import { findResourceById } from '../db/DBOperations';
 import { Service } from '../types/service';
+import { ResourceNotFoundError } from '../util/errorUtils';
```
mgramigna, line 4: Guess this is no longer an example now that we have something huh?

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+let server: Server;
+
+const Library: fhir4.Library = {
+  resourceType: 'Library',
+  type: { coding: [{ code: 'logic-library' }] },
+  id: 'library123',
+  status: 'active'
+};
+
+describe('LibraryService', () => {
```
mgramigna, line 15: Seems like these were mostly copied from deqm-test-server, which is fine, but can we get a fresh start on some of the formatting? Specifically the `it` over `test` and test naming conventions here: https://gist.github.com/mgramigna/bdc933fae2150216681c28d1f0f7aac7#testing

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
+    await testSetup([Library]);
+  });
```
mgramigna, line 20:
```suggestion
  });
  
```

As a selfish vim user, it's really hard to navigate tests when there aren't spaces between the different blocks. In general I think it helps the readability overall though

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    test('test searchById with correctHeaders and the id should be in database returns 200', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library/library123')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then((response: any) => {
+          expect(response.body.id).toEqual(Library.id);
+        });
+    });
```
mgramigna, line 31:
```suggestion
    });
    
```

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        .set('Accept', 'application/json+fhir')
+        .expect(404)
+        .then((response: any) => {
+          expect(response.body.issue[0].code).toEqual('ResourceNotFound');
+          expect(response.body.issue[0].details.text).toEqual(
+            `No resource found in collection: Library, with: id invalidID`
+          );
+        });
+    });
+  });
```
mgramigna, line 44:
```suggestion
  });
  
```

test/MeasureService.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const Measure: fhir4.Measure = { resourceType: 'Measure', id: 'measure123', status: 'active' };
+
+describe('measure.service', () => {
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
+
+    await testSetup([Measure]);
+  });
```
mgramigna, line 16:
```suggestion
  });
  
```

test/MeasureService.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    test('test searchById with correctHeaders and the id should be in database returns 200', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Measure/measure123')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then((response: any) => {
+          expect(response.body.id).toEqual(Measure.id);
+        });
+    });
```
mgramigna, line 27:
```suggestion
    });
    
```

test/MeasureService.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        .set('Accept', 'application/json+fhir')
+        .expect(404)
+        .then((response: any) => {
+          expect(response.body.issue[0].code).toEqual('ResourceNotFound');
+          expect(response.body.issue[0].details.text).toEqual(
+            `No resource found in collection: Measure, with: id invalidID`
+          );
+        });
+    });
+  });
```
mgramigna, line 40:
```suggestion
  });
  
```

test/MeasureService.test.ts
```diff
@@ -0,0 +1,42 @@
+import { initialize, Server } from '@projecttacoma/node-fhir-server-core';
+import { serverConfig } from '../src/config/serverConfig';
+import { cleanUpDb, testSetup } from './utils';
+const supertest = require('supertest');
+
+let server: Server;
+
+const Measure: fhir4.Measure = { resourceType: 'Measure', id: 'measure123', status: 'active' };
+
+describe('measure.service', () => {
```
mgramigna, line 10: Same comments about the test styling here. Also, the describe is named `measure.service` here but `LibraryService` in the library service. Can we be consistent?

test/LibraryService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  status: 'active'
+};
+
+describe('LibraryService', () => {
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
+    await testSetup([Library]);
+  });
+  describe('CRUD operations', () => {
```
mgramigna, line 21: Idk if it makes sense to group this test as `CRUD operations`. Kinda misleading since this server is probably only going to support the `R` part of `CRUD`. Maybe we rename this describe block to just be the function that it's testing: `searchById`?

```suggestion
  describe('searchById', () => {
```

test/MeasureService.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
+
+    await testSetup([Measure]);
+  });
+  describe('CRUD operations', () => {
```
mgramigna, line 17: See above

```suggestion
  describe('searchById', () => {
```

src/services/MeasureService.ts
```diff
@@ -12,10 +15,16 @@ export class MeasureService implements Service<fhir4.Measure> {
...
-  // @ts-ignore
-  async searchById() {
-    throw new Error('Not implemented');
+  /**
+   *
+   * result of sending a GET request to {BASE_URL}/4_0_1/Measure/{id}
+   * searches for the measure with the passed in id
+   */
+  async searchById(args: RequestArgs) {
+    const result = await findResourceById(args.id, 'Measure');
```
mgramigna, line 24: Thoughts on logging? We should probably log something about when each service function is called, maybe something like:

```ts
logger.info(`GET /Measure/${args.id}`);
```

requires importing the logger in this file. See `index.ts` for an example of that

src/services/LibraryService.ts
```diff
@@ -0,0 +1,29 @@
...
+  // @ts-ignore
+  async search() {
+    throw new Error('Not implemented');
+  }
+
+  /**
+   * result of sending a GET request to {BASE_URL}/4_0_1/Library/{id}
+   * searches for the library with the passed in id
+   */
+  async searchById(args: RequestArgs) {
```
mgramigna, line 22: Same thoughts on logging here

mgramigna (changes requested):
Works great.

I have a lot of code styling comments. Now that we're starting fresh with this, I understand how copying a lot of things from deqm-test-server will be useful. I think it's totally fine to do that, but we should make sure that we take the opportunity to clean things up where appropriate, because a lot of that code has seen better days at this point.

test/utils.ts
elsaperelli: Hahahah

src/services/MeasureService.ts
sarahmcdougall: ^ I agree - I think this would be a good place to do some logging, and Matt's example seems good since it's straight to the point

src/services/LibraryService.ts
elsaperelli: 😮

sarahmcdougall (changes requested):
I agree with all Matt's comments, plus added a few more suggestions!

test/LibraryService.test.ts
elsaperelli: Oh yes how quickly I forgot about that...good thing we have the doc

src/index.ts
natjoe4: I thought that if we import Connection here, and the envConfig file was imported in Connection it would actually get run before this line is executed... is that not the case? Or, by far up, do you mean as close to the code that starts the server as possible. This was also convenient for testing purposes since we could just import Connection in our test utils file and get envConfig to run along with it, but that can be changed by just importing the envConfig there if you think it makes sense to keep `dotenv.config()` in this file.

src/db/DBOperations.ts
```diff
@@ -0,0 +1,10 @@
+import { FhirResourceType } from '@projecttacoma/node-fhir-server-core';
```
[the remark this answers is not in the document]
elsaperelli: Ah yes good catch!

test/LibraryService.test.ts
```diff
@@ -0,0 +1,49 @@
+import { initialize, Server } from '@projecttacoma/node-fhir-server-core';
```
mgramigna, line 1: Forgot this comment , but let's move `LibraryService.test.ts` and `MeasureService.test.ts` to  a subdirectory `test/services/` for cleanliness

src/index.ts
mgramigna: Bonus: we can also silence the logger in this file as shown in the example.

I think I stole this code from you, Nat. Lol

test/services/MeasureService.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    await testSetup([Measure]);
+  });
+
+  describe('searchById', () => {
+    it('test searchById with correctHeaders and the id should be in database returns 200', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Measure/measure123')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
```
mgramigna, line 23: Nitpicking here, but this content-type header actually does nothing.

It matters in the request when we are specifying the content-type of the request body, which is impossible in GET requests.

It will matter if we invoke post requests with FHIR resources as the body


```suggestion
```

test/services/LibraryService.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    server = initialize(config);
+    await testSetup([Library]);
+  });
+
+  describe('searchById', () => {
+    it('test searchById with correctHeaders and the id should be in database returns 200', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library/library123')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
```
mgramigna, line 27: Nitpicking here, but this content-type header actually does nothing.

It matters in the request when we are specifying the content-type of the request body, which is impossible in GET requests.

It will matter if we invoke post requests with FHIR resources as the body

```suggestion
```

src/services/LibraryService.ts
```diff
@@ -0,0 +1,32 @@
...
+    const result = await findResourceById(args.id, 'Library');
+    logger.info(`GET /Library/${args.id}`);
```
mgramigna, line 26:
```suggestion
    logger.info(`GET /Library/${args.id}`);
    const result = await findResourceById(args.id, 'Library');
```

sorry, lol

src/services/MeasureService.ts
```diff
@@ -12,10 +17,16 @@ export class MeasureService implements Service<fhir4.Measure> {
...
-  // @ts-ignore
-  async searchById() {
-    throw new Error('Not implemented');
+  /**
+   * result of sending a GET request to {BASE_URL}/4_0_1/Measure/{id}
+   * searches for the measure with the passed in id
+   */
+  async searchById(args: RequestArgs) {
+    const result = await findResourceById(args.id, 'Measure');
+    logger.info(`GET /Measure/${args.id}`);
```
mgramigna, line 26: Sorry, lol

```suggestion
    logger.info(`GET /Measure/${args.id}`);
    const result = await findResourceById(args.id, 'Measure');
```

mgramigna (changes requested):

test/services/MeasureService.test.ts
```diff
@@ -0,0 +1,44 @@
...
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('ResourceNotFound');
+          expect(response.body.issue[0].details.text).toEqual(
+            `No resource found in collection: Measure, with: id invalidID`
+          );
+        });
+    });
+  });
+
+  afterAll(cleanUpDb);
```
mgramigna, line 43: I was curious about the usage of afterAll here and how a reference to the function is passed in.

I'm not actually sure if this makes a difference at all, but I was looking at the [Jest docs for setup and teardown](https://jestjs.io/docs/setup-teardown#one-time-setup) and it seems like the example there is to return the actual promise from these hooks when the operations are asynchronous. Maybe we could change it to this approach? It could potentially help us avoid database race condition issues in tests that we've seen before in deqm-test-server, idk. This would make it look like this (and would apply to LibraryService.test.ts as well:


```ts
  beforeAll(() => {
    const config = serverConfig;
    server = initialize(config);
    return testSetup([Library]);
  });

// ...

  afterAll(() => {
    return cleanUpDb();
  });

```

test/services/LibraryService.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  resourceType: 'Library',
+  type: { coding: [{ code: 'logic-library' }] },
+  id: 'library123',
+  status: 'active'
+};
+
+describe('LibraryService', () => {
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
```
mgramigna, line 18: Also, is this really necessary?

```suggestion
    server = initialize(serverConfig);
```

test/services/MeasureService.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import supertest from 'supertest';
+
+let server: Server;
+
+const Measure: fhir4.Measure = { resourceType: 'Measure', id: 'measure123', status: 'active' };
+
+describe('MeasureService', () => {
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
```
mgramigna, line 13: Also, is this really necessary?

```suggestion
    server = initialize(serverConfig);
```

mgramigna (changes requested):

mgramigna (changes requested):

test/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
+  await collection.insertOne({ ...data });
+};
+
+export async function cleanUpDb() {
+  await Connection.db.dropDatabase();
+  await Connection.connection?.close();
+}
+
+export async function testSetup(testFixtureList: fhir4.FhirResource[]) {
```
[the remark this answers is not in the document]
sarahmcdougall: ^ I agree, `testDatabaseSetup` sounds good, or maybe `setupTestDatabase` (personally when I see `testSetup`/`testDatabaseSetup` I associate the function with testing the setup rather than getting set up for the tests so I prefer the verb coming first but that's just me)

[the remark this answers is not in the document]
elsaperelli: `testDatabaseSetup` sounds nice :)

mgramigna (approved):
Great work, both of you. Really happy with where this PR ended up

sarahmcdougall (approved):
Everything looks good to me! Very exciting to be starting up the measure repository service :)

PR #3, opened 2022-12-12
Search (an Elsa Perelli and Nat Jones colab)

natjoe4:
## Summary
Measure Repository Service now supports search!

### New Behavior
The endpoints: 
 - `GET` to `/4_0_1/Library?{QUERY}`
 and
 - `GET` to `/4_0_1/Measure?{QUERY}`
 are now active and return FHIR searchset bundles containing all resources on the server of the desired resourceType that match the included query criteria!
### Code Changes
 - Implemented `search` function in both MeasureService and LibraryService
 - Created `findResourcesWithQuery` function in `dbOperations.ts`
 - Created `validateSearchParams` function in `validationUtils.ts`
 - Created `createSearchsetBundle` function in `bundleUtils.ts`
 - Updating jest config to include `@shelf/jest-mongodb`, which allows us to easily use in-memory mongo databases for unit tests
 - Created `jest-mongodb-config.js` in line with `@shelf/jest-mongodb` specifications
 - Installed `uuid` as a dependency for bundle generation
 - Added `check` script to package.json
## Testing Guidance
- ~`npm run test`~ *try the new and improved* `npm run check`
- ~`npm run lint`~ 🥱 old news
- `npm run db:reset`
- on line 21 in `index.ts`, add the following code snippet: 
```
  .then(() => {
    Connection.db.collection('Measure').insertOne({
      resourceType: 'Measure',
      id: 'Measure1',
      url: 'http://example.com',
      status: 'active',
      version: 'testVersion',
      name: 'testName',
      identifier: { system: 'testSystem', value: 'testValue' },
      title: 'testTitle',
      description: 'testDescription'
    });
    Connection.db
      .collection('Measure')
      .insertOne({ resourceType: 'Measure', id: 'Measure2', url: 'http://example2.com', status: 'active' });
    Connection.db
      .collection('Library')
      .insertOne({
        resourceType: 'Library',
        id: 'Library1',
        url: 'http://example.com',
        status: 'active',
        version: 'testVersion',
        name: 'testName',
        identifier: { system: 'testSystem', value: 'testValue' },
        title: 'testTitle',
        description: 'testDescription'
      });
    Connection.db
      .collection('Library')
      .insertOne({ resourceType: 'Library', id: 'Library2', url: 'http://example2.com', status: 'active' });
  })
```
 - Import the uploaded document into insomnia. For instructions on how to do this, reference [this](https://confluence.mitre.org/display/TAC/Sample+Requests)
 - `npm start`
 - Try submitting each of the requests and ensure that the output is as expected!
 - Please let us know if you have any questions!
[Search_Testing.json.zip](https://github.com/projecttacoma/measure-repository-service/files/10211756/Search_Testing.json.zip)


test/services/LibraryService.test.ts
```diff
@@ -5,31 +5,40 @@ import supertest from 'supertest';
...
 };
 
 describe('LibraryService', () => {
-  beforeAll(() => {
-    server = initialize(serverConfig);
-    return setupTestDatabase([Library]);
+  beforeAll(async () => {
+    const config = serverConfig;
+    server = initialize(config);
+    await setupTestDatabase([LIBRARY, LIBRARY_WITH_URL]);
```
mgramigna, line 27: Looks like some of the requested changes from #2 were lost in translation during a rebase. Can we follow the pattern suggested in those changes instead?

```suggestion
  beforeAll(() => {
    server = initialize(serverConfig);
    return setupTestDatabase([LIBRARY, LIBRARY_WITH_URL]);
```

test/services/LibraryService.test.ts
```diff
@@ -5,31 +5,40 @@ import supertest from 'supertest';
...
   });
 
   describe('searchById', () => {
-    it('test searchById with correctHeaders and the id should be in database returns 200', async () => {
+    it('returns 200 when passed correct headers and the id is in database ', async () => {
```
mgramigna, line 31: I appreciate the phrasing change here and in others :)

test/services/LibraryService.test.ts
```diff
@@ -42,7 +51,38 @@ describe('LibraryService', () => {
...
+    it('returns 200 and correct searchset bundle when query matches multiple resources', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library')
+        .query({ status: 'active' })
+        .set('Accept', 'application/json+fhir')
+        .expect(200)
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Bundle');
+          expect(response.body.total).toEqual(2);
+          expect(response.body.entry.find((e: fhir4.BundleEntry) => e.resource?.id === 'test')?.resource).toEqual(
```
mgramigna, line 77: This is not a change request, but what do you think of this assertion style instead?

```ts
expect(response.body.entry).toEqual(
  expect.arrayContaining([
    expect.objectContaining<fhir4.BundleEntry>({
      resource: LIBRARY
    }),
    expect.objectContaining<fhir4.BundleEntry>({
      resource: LIBRARY_WITH_URL
    })
  ])
);
```

Totally cool if you prefer the .find, I'll leave it up to you

src/util/queryUtils.ts
```diff
@@ -0,0 +1,24 @@
+import { RequestQuery } from '@projecttacoma/node-fhir-server-core';
+
+/**
+ * Takes in an express query and parses the identifier field, if present, into
+ * a usable mongo query
+ */
+export function parseQuery(query: RequestQuery): RequestQuery {
+  if (query.identifier) {
+    const iden = query.identifier;
+    delete query.identifier;
```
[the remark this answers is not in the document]
mgramigna: Matt from the future here. I realized by looking more at the code that this issue is not as simple as I thought, and that the reason query is modified is to allow the existing req.query entries to be preserved when doing the mongo query.

I think my comment is still valuable so I'm not going to delete it, but maybe we can discuss a compromise appproach to not just pass `req.query` directly through the mongo after passing the validation. I wonder if there's a way we can be more explicit or something. Maybe not, but we'll see

mgramigna (changes requested):
Everything works great. I have a bunch of inline comments, and the following:

I found something that I find pretty interesting. Specifically, I think handling of search params that are a string type (i.e. name, title, description), need special handling. From the [FHIR search docs](https://build.fhir.org/search.html#ptypes), the table entry for `string` type:

> Search parameter is a simple string, like a name part. Search is case-insensitive and accent-insensitive. **May match just the start of a string**. String parameters may contain spaces.

This means that for these string type parameters, we need to construct the mongo query in a way such that it can match the start of a string rather than the full thing, e.g.:

`http://localhost:3000/4_0_1/Measure?description=testDescription` and `http://localhost:3000/4_0_1/Measure?description=test` Should both work.

src/util/queryUtils.ts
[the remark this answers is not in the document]
mgramigna: Matt from the future again. After realizing my top-level comment about string searches, it seems inevitable that the approach of passing req.query through will have to change

mgramigna (approved):
Looks good, might have to consider pagination in the future depending on the quantities of measures that we're dealing with. For another day though.

PR #4, opened 2022-12-19
Readme updates

natjoe4:
# Summary
Updated the `README.md` to reflect the recent changes to functionality. Also added `start:clean` script which can be run from command line.
 
## New behavior

- `README.md` updates to reflect `readById` and `search` endpoints as well as instructions for running with mongo and insomnia
- Added `npm run start:clean` script which runs `db:reset` and `start` scripts in sequence for a clean start

## Code changes

- `README.md` updates
- `package.json` script addition

# Testing guidance
 - View the `README.md` file in a markdown text previewer and make sure all information is accurate and no information is missing
 - Test the `npm run start:clean` script by ading code to `index.ts` to create resources on the server at server start as outlined in #3. Then, remove this code and `ctrl + c` the server.  then run `npm run start:clean` and use Robo3T to check that the server has started with a clean db :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,19 @@ A prototype implementation of a [FHIR Measure Repository Service](https://build.
 - [Node.js >=16.0.0](https://nodejs.org/en/)
 - [MongoDB >= 6.0](https://www.mongodb.com)
 
+### MongoDB
+
+This test server makes use of [MongoDB](https://www.mongodb.com), a cross-platform document-oriented database program.
+
+Follow the [MongoDB Community Edition installation guide](https://docs.mongodb.com/manual/installation/) for your platform, and follow the commands for running MongoDB on your machine.
+
 ## Usage
 
-```
-npm start
-```
+Once MongoDB is running on your machine, run the `npm start` command to start up the FHIR server at `localhost:3000`.
+
+For ease of testing, it is recommended to download [Insomnia API Client and Design Tool](https://insomnia.rest) for sending HTTP requests to the server and [Robo 3T](https://robomongo.org) as a GUI for viewing the Mongo database.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,6 +30,17 @@ To create these collections:
...
+### CRUD Operations
+
+This server currently supports the following CRUD operations:
+
+- Read by ID to endpoint: `4_0_1/<resourceType>/<resourceId>`
+  _More functionality coming soon!_
+
+### Search
+
+The Measure Repository Server supports `Library` and `Measure` resource search by the parameters specified in the shareable measure repository section of the [HL7 Measure Repository Docs](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#shareable-measure-repository).
```
elsaperelli, line 42: I could be very wrong by this but do we want to refer to this as the Measure Repository Service Server or just Measure Repository Service? Or is Measure Repository Server also correct? (sounds right, just wondering)

elsaperelli, line 42: Do we want to list the parameters here so the user doesn't have to navigate to the docs? Might not be necessary I just saw that we do that in the test-server README.

README.md
```diff
@@ -22,6 +30,17 @@ To create these collections:
 npm run db:setup
```
elsaperelli, line 30: Maybe add the new `npm run start:clean` that you just added as well?

elsaperelli (changes requested):
Just a few suggestions! Nice addition of `start:clean`!

README.md
natjoe4: Good catch! This was accidental. I think Measure Repository Service Server is probably what we want, although it is a mouthful :/

elsaperelli (approved):
Lgtm!

mgramigna (approved):

PR #5, opened 2022-12-21
Measure $package

natjoe4:
# Summary
Creates the `/4_0_1/Measure/$package` endpoint which bundles a Measure with all its Libraries and dependent Libraries recursively, as outlined in the [docs](https://build.fhir.org/ig/HL7/cqf-measures/OperationDefinition-Measure-package.html )

## New behavior
 - 4 new endpoints:
 - `GET 4_0_1/Measure/$package`
 - `GET 4_0_1/Measure/<id>/$package`
 - `POST 4_0_1/Measure/$package`
 - `POST 4_0_1/Measure/<id>/$package`
 - Wowza new `db:loadBundle` script!

## Code changes
 - Added new endpoints in `serverConfig`
 - Created `db:loadBundle` script
 - lots n' lots of testing

# Testing guidance
 - `npm run check`
 - `npm run db:reset`
 - ```npm run db:loadBundle "ecqm-content-r4-2021/bundles/measure/ColorectalCancerScreeningsFHIR/ColorectalCancerScreeningsFHIR-bundle.json"```
 - `npm start`
 - Run through the included Insomnia requests and make sure all outcomes are as expected!
 - For all passing requests, we expect the Bundle to include the Measure itself as well as 9 Library resources. The data will be too large for Insomnia to show it by default so you'll need to just click show anyway. The expected returned libraries are: 
 - `ColorectalCancerScreeningsFHIR`
 - `FHIRHelpers`
 - `SupplementalDataElementsFHIR4`
 - `MATGlobalCommonFunctionsFHIR4`
 - `AdultOutpatientEncountersFHIR4`
 - `HospiceFHIR4`
 - `AdvancedIllnessandFrailtyExclusionECQMFHIR4`
 - `CumulativeMedicationDurationFHIR4`
 - `PalliativeCareFHIR`
 [Measure-$package-testing.json.zip](https://github.com/projecttacoma/measure-repository-service/files/10281840/Measure-.package-testing.json.zip)

### DIFF
--- a/scripts/dbSetup.ts
+++ b/scripts/dbSetup.ts
@@ -1,4 +1,7 @@
 import { Connection } from '../src/db/Connection';
+import * as fs from 'fs';
+import { Bundle, FhirResource } from 'fhir/r4';
+import { MongoError, OptionalId } from 'mongodb';
 
 const DB_URL = process.env.DATABASE_URL || 'mongodb://localhost:27017/measure-repository';
 const COLLECTION_NAMES = ['Measure', 'Library', 'MeasureReport'];
@@ -27,6 +30,66 @@ async function deleteCollections() {
   await Promise.all(deletions);
 }
 
+/*
+ * Loads all measure or library resources found in the bundle located at param filePath
+ */
+async function loadBundle(filePath: string) {
+  await Connection.connect(DB_URL);
+  console.log(`Connected to ${DB_URL}`);
+  console.log(`Loading bundle from path ${filePath}`);
+
+  const data = fs.readFileSync(filePath, 'utf8');
+  if (data) {
+    console.log(`Uploading ${filePath.split('/').slice(-1)}...`);
+    const bundle: Bundle = JSON.parse(data);
+    // retrieve each resource and insert into database
+    if (bundle.entry) {
+      let resourcesUploaded = 0;
+      const uploads = bundle.entry.map(async res => {
+        // Only upload Library or Measure resources
+        if (
+          res?.resource?.resourceType &&
+          (res?.resource?.resourceType === 'Library' || res?.resource?.resourceType === 'Measure')
+        ) {
+          try {
+            await createResource(res.resource, res.resource.resourceType);
+            resourcesUploaded += 1;
+          } catch (e) {
+            // ignore duplicate key errors for Libraries, Note: if ValueSets added, also ignore for those
+            if (!(e instanceof MongoError && e.code == 11000 && res.resource.resourceType === 'Library')) {
+              if (e instanceof Error) {
+                console.error(e.message);
+              } else {
+                console.error(String(e));
+              }
+            }
+          }
+        } else {
+          console.log(
+            res?.resource?.resourceType
+              ? `Not loading resource of type ${res?.resource?.resourceType}`
+              : 'Resource or resource type undefined'
+          );
+        }
+      });
+      await Promise.all(uploads);
+      console.log(`${resourcesUploaded} resources uploaded.`);
+    } else {
+      console.error('Unable to identify bundle entries.');
+    }
+  }
+}
+
+/*
+ * Inserts one data object into database with specified FHIR resource type
+ */
+async function createResource(data: FhirResource, resourceType: string) {
+  const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
+  console.log(`Inserting ${resourceType}/${data.id} into database`);
+  await collection.insertOne(data);
+  return { id: data.id };
+}
+
 if (process.argv[2] === 'delete') {
   deleteCollections()
     .then(() => {
@@ -50,6 +113,18 @@ if (process.argv[2] === 'delete') {
     .then(() => {
       return createCollections();
     })
+    .then(() => {
+      console.log('Done');
+    })
+    .catch(console.error)
+    .finally(() => {
+      Connection.connection?.close();
+    });
+} else if (process.argv[2] === 'loadBundle') {
+  if (process.argv.length < 4) {
+    throw new Error('Filename argument required.');
+  }
+  loadBundle(process.argv[3])
     .then(() => {
       console.log('Done');
     })
--- a/scripts/dbSetup.ts
+++ b/scripts/dbSetup.ts
@@ -1,7 +1,6 @@
 import { Connection } from '../src/db/Connection';
 import * as fs from 'fs';
-import { Bundle, FhirResource } from 'fhir/r4';
-import { MongoError, OptionalId } from 'mongodb';
+import { MongoError } from 'mongodb';
 
 const DB_URL = process.env.DATABASE_URL || 'mongodb://localhost:27017/measure-repository';
 const COLLECTION_NAMES = ['Measure', 'Library', 'MeasureReport'];
@@ -41,7 +40,7 @@ async function loadBundle(filePath: string) {
   const data = fs.readFileSync(filePath, 'utf8');
   if (data) {
     console.log(`Uploading ${filePath.split('/').slice(-1)}...`);
-    const bundle: Bundle = JSON.parse(data);
+    const bundle: fhir4.Bundle = JSON.parse(data);
     // retrieve each resource and insert into database
     if (bundle.entry) {
       let resourcesUploaded = 0;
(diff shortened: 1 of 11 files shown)


src/util/bundleUtils.ts
```diff
@@ -14,3 +20,90 @@ export function createSearchsetBundle<T extends fhir4.FhirResource>(entries: T[]
...
+  }
+  // This filter checks for the 'Library' keyword on all related artifacts
+  const depLibUrls = lib.relatedArtifact
+    .filter(
+      ra =>
+        ra.type === 'depends-on' &&
+        ra.resource?.includes('Library') &&
+        ra.resource !== 'http://fhir.org/guides/cqf/common/Library/FHIR-ModelInfo|4.0.1'
+    ) // exclude modelinfo dependency
+    .map(ra => ra.resource as string); //TODO: may be able to improve this typing
```
mgramigna, line 89: Not a reviewer but just saying that the casting to string is fine here. You know more than typescript in the sense that anything where `ra.resource` is undefined will have already been filtered out of the array so it's safe to always do `as string`. TypeScript isn't smart enough to recognize the  chaining of the array calls and how they influence each other in that regard. Can probably remove the TODO

src/util/bundleUtils.ts
```diff
@@ -14,3 +20,90 @@ export function createSearchsetBundle<T extends fhir4.FhirResource>(entries: T[]
...
+    return result;
+  } else {
+    throw new BadRequestError(`Uploaded measure: ${measure.id} does not reference a Library`);
+  }
+}
+
+/**
+ * Assemble a mongo query based on a reference to another resource
+ * @param {string} reference assumed to be canonical
+ * @returns {Filter} mongo query to pass in to mongo controller to search for the referenced resource
```
elsaperelli, line 59: We don't necessarily need these since we are in TypeScript, right? Although I do think it's good to have that the string param is assumed to be canonical so we can either keep it or just include it in the overall description.

elsaperelli (changes requested):
Awesome work Lauren/Nat! It all looks great to me- just one small comment!

src/util/bundleUtils.ts
mgramigna: Yeah I've come around to the take of basically it's okay when some stricter assumption needs to be clarified about something.

Like if it was just `@param {string} reference the reference` I'd say delete, but this is probably fine

src/util/validationUtils.ts
```diff
@@ -13,3 +13,26 @@ export function validateSearchParams(query: RequestQuery) {
     throw new BadRequestError(`Parameters ${invalidParams.join(', ')} are not valid for search`);
   }
 }
+
+/*
+ * Gathers parameters from both the query and the FHIR parameter request body resource
+ */
+export function gatherParams(query: RequestQuery, parameters?: fhir4.Parameters) {
+  const gatheredParams: Record<string, any> = { ...query };
```
sarahmcdougall, line 21: Thoughts on replacing the `any`? I think the correct typing here would be `string | boolean | undefined | number | string[]` which is a mouthful 😅

sarahmcdougall (changes requested):
Good work with this :) It's nice that we were able to port over some code from the test server for this! Aside from in-line comments, a few things:

- I noticed the PR description and testing guidance lists the script name as `db:loadBundles` instead of `db:loadBundle`. Really minor but I think we should update the PR description so that it's accurate, just in case anyone takes a look at it in the future (or for anyone like me who copy/pastes the testing commands to avoid having to type long file paths :no_mouth:)
- Not sure of the answer to this - does it make sense to include the load bundle functionality in `dbSetup.ts`, or should we separate it into a new file? My thought here is that loading bundles is an act of populating the database with data, but not necessarily "setting it up" in the sense of creating collections. At the same time, I don't want this repo to get flooded with a bunch of files, so not sure what the "correct" option is here... just something that came across my mind as I was reviewing

src/util/bundleUtils.ts
lmd59: Thanks! Removed TODO

lmd59:
> Good work with this :) It's nice that we were able to port over some code from the test server for this! Aside from in-line comments, a few things:
> 
> * I noticed the PR description and testing guidance lists the script name as `db:loadBundles` instead of `db:loadBundle`. Really minor but I think we should update the PR description so that it's accurate, just in case anyone takes a look at it in the future (or for anyone like me who copy/pastes the testing commands to avoid having to type long file paths 😶)
> * Not sure of the answer to this - does it make sense to include the load bundle functionality in `dbSetup.ts`, or should we separate it into a new file? My thought here is that loading bundles is an act of populating the database with data, but not necessarily "setting it up" in the sense of creating collections. At the same time, I don't want this repo to get flooded with a bunch of files, so not sure what the "correct" option is here... just something that came across my mind as I was reviewing

Fixed point one. On the second point, I could see moving it based on "creating collections" sense, but it does seem to fit into a more general category of database setup in which you likely need *some* data to play with to do anything of interest. I could go either way, but lean toward as-is because easier :-P

scripts/dbSetup.ts
```diff
@@ -27,6 +29,66 @@ async function deleteCollections() {
...
+                console.error(String(e));
+              }
+            }
+          }
+        } else {
+          console.log(
+            res?.resource?.resourceType
+              ? `Not loading resource of type ${res?.resource?.resourceType}`
+              : 'Resource or resource type undefined'
+          );
```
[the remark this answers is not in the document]
lmd59: Good suggestion!

src/config/serverConfig.ts
```diff
@@ -5,7 +5,33 @@ export const serverConfig: ServerConfig = {
...
     Measure: {
       service: new MeasureService(),
-      versions: [constants.VERSIONS['4_0_1']]
+      versions: [constants.VERSIONS['4_0_1']],
+      operation: [
+        {
+          name: 'package',
+          route: '/$package',
+          method: 'GET',
+          reference: 'https://build.fhir.org/ig/HL7/cqf-measures/OperationDefinition-Measure-package.html'
```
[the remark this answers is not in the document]
lmd59: When I go to the URL you've linked here, it says:

> The official URL for this operation definition is:

> http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Measure-package

So I'm going to go with that instead of including the "STU3" bit if that works!

src/services/MeasureService.ts
```diff
@@ -39,4 +40,45 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+
+    // query construction
+    const query: Filter<any> = {};
+    if (id) query.id = id;
+    if (url) query.url = url;
+    if (version) query.version = version;
+    const measure = await findResourcesWithQuery<fhir4.Measure>(query, 'Measure');
+    if (!measure || !(measure.length > 0)) {
+      throw new ResourceNotFoundError(
+        `No resource found in collection: Measure, with: ${Object.keys(query)
```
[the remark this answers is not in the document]
lmd59: This is actually a bit weird in a couple of places, where the colon comes after with instead of the thing being identified. Changed up across the board.

src/util/validationUtils.ts
lmd59: Hmmm- this is a tricky one. Currently, `gatherParams` is meant to handle anything that we'd throw at it with `$package`, but we're pretty likely to expand on the kinds of params that we'd gather in the future. When I look at the valueX options in `ParametersParameter` type, this list does cover most of the common ones, but there are plenty of other types that could come in as parameters on some operation in the future. I would lean toward keeping it `any` as it could capture all of that weird stuff as well as needed (and this function's stated purpose is to collect anything we need from a query). But I could be convinced otherwise if we think the `any` here mitigates enough other risk.

elsaperelli (approved):
Great job Lauren & Nat! I agree with all the changes made from Sarah's comments. Lgtm.

sarahmcdougall:
> On the second point, I could see moving it based on "creating collections" sense, but it does seem to fit into a more general category of database setup in which you likely need _some_ data to play with to do anything of interest. I could go either way, but lean toward as-is because easier :-P

Sounds good to me - I figure we probably won't need many more scripts in this repo anyway, so it's fine to just keep everything in that one file 👍

sarahmcdougall (approved):
All the changes look good! I like the new setup of saying XX resources were skipped in the `db:loadBundle` script - looks clean but still informative 👍

PR #6, opened 2022-12-29
Library $package 

elsaperelli:
# Summary
Creates the `/4_0_1/Library/$package` endpoint which bundles a Library with all of its dependent libraries recursively, as outlined in the [docs](https://build.fhir.org/ig/HL7/cqf-measures/OperationDefinition-Library-package.html).

## New behavior
- 4 new endpoints:
- `GET 4_0_1/Library/$package`
- `GET 4_0_1/Library/<id>/$package`
- `POST 4_0_1/Library/$package`
- `POST 4_0_1/Library/<id>/$package`
- Additional testing for these endpoints as well as some for Measure/$package

## Code changes
- `src/config/serverConfig.ts` - added above endpoints
- `src/services/LibraryService.ts` - added `package` function which handles the above endpoints. It is similar to the `package` function in `MeasureService.ts`.
- `src/util/bundleUtils` - created a `createLibraryPackageBundle` function and refactored the reusable code in `createMeasurePackageBundle` into a new function called `createDepLibraryBundle` so that it can be used by both services.
- `test/services/LibraryService.test.ts` - added tests 
- `test/services/MeasureService.test.ts` - I noticed that one of the test descriptions was misleading so I changed it and then added two additional tests. Some of these tests may be redundant so let me know if some of these can be removed (same with tests in `LibraryService.test.ts`.
- `test/util/bundleUtils.test.ts` - added tests for `createLibraryPackageBundle` function and `createDepLibraryBundle` function.

# Testing guidance
_Inspired by Testing guidance by Nat/Lauren in the [Measure $package PR](https://github.com/projecttacoma/measure-repository-service/pull/5)_ 
- `npm run check`
- `npm run db:reset`
- `npm start`
- Run through the included Insomnia requests and make sure all outcomes are as expected!
- For all passing requests, we expect the Bundle to include 9 Library resources. The data will be too large for Insomnia to show it by default so you'll need to just click show anyway. **NOTE:** this is the same outcome as the `Measure/$package` operation, but without the Measure.
- The expected returned libraries are:
- `ColorectalCancerScreeningsFHIR`
- `FHIRHelpers`
- `SupplementalDataElementsFHIR4`
- `MATGlobalCommonFunctionsFHIR4`
- `AdultOutpatientEncountersFHIR4`
- `HospiceFHIR4`
- `AdvancedIllnessFrailtyExclusionECQMFHIR4`
- `CumulativeMedicationDurationFHIR4`
- `PalliativeCareFHIR`

[Library-$package-testing.json.zip](https://github.com/projecttacoma/measure-repository-service/files/10322230/Library-.package-testing.json.zip)

### DIFF
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -45,7 +45,7 @@ export class LibraryService implements Service<fhir4.Library> {
    * result of sending a POST or GET request to:
    * {BASE_URL}/4_0_1/Library/$package or {BASE_URL}/4_0_1/Library/:id/$package
    * creates a bundle of the library (specified by parameters) and all dependent libraries
-   * requires parameters id and/or url, but also supports version as supplemental (optional)
+   * requires parameters id, url, and/or identifier, but also supports version as supplemental (optional)
    */
   async package(args: RequestArgs, { req }: RequestCtx) {
     logger.info(`${req.method} ${req.path}`);
@@ -57,10 +57,9 @@ export class LibraryService implements Service<fhir4.Library> {
     const identifier = params.identifier;
 
     if (!id && !url && !identifier) {
-      throw new BadRequestError('Must provide identifying information via either id or url parameters');
+      throw new BadRequestError('Must provide identifying information via either id, url, or identifier parameters');
     }
 
-    // query construction
     const query: Filter<any> = {};
     if (id) query.id = id;
     if (url) query.url = url;
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -54,8 +54,9 @@ export class LibraryService implements Service<fhir4.Library> {
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
+    const identifier = params.identifier;
 
-    if (!id && !url) {
+    if (!id && !url && !identifier) {
       throw new BadRequestError('Must provide identifying information via either id or url parameters');
     }
 
@@ -64,7 +65,10 @@ export class LibraryService implements Service<fhir4.Library> {
     if (id) query.id = id;
     if (url) query.url = url;
     if (version) query.version = version;
-    const library = await findResourcesWithQuery<fhir4.Library>(query, 'Library');
+    if (identifier) query.identifier = identifier;
+
+    const parsedQuery = getMongoQueryFromRequest(query);
+    const library = await findResourcesWithQuery<fhir4.Library>(parsedQuery, 'Library');
     if (!library || !(library.length > 0)) {
       throw new ResourceNotFoundError(
         `No resource found in collection: Library, with ${Object.keys(query)
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -1,10 +1,11 @@
 import { loggers, RequestArgs, RequestCtx } from '@projecttacoma/node-fhir-server-core';
+import { Filter } from 'mongodb';
 import { findResourceById, findResourcesWithQuery } from '../db/dbOperations';
 import { Service } from '../types/service';
-import { createSearchsetBundle } from '../util/bundleUtils';
-import { ResourceNotFoundError } from '../util/errorUtils';
+import { createLibraryPackageBundle, createSearchsetBundle } from '../util/bundleUtils';
+import { BadRequestError, ResourceNotFoundError } from '../util/errorUtils';
 import { getMongoQueryFromRequest } from '../util/queryUtils';
-import { validateSearchParams } from '../util/validationUtils';
+import { gatherParams, validateSearchParams } from '../util/validationUtils';
 
 const logger = loggers.get('default');
 
@@ -38,5 +39,47 @@ export class LibraryService implements Service<fhir4.Library> {
       throw new ResourceNotFoundError(`No resource found in collection: Library, with id: ${args.id}`);
     }
     return result;
+  }
+
+  /**
+   * result of sending a POST or GET request to:
+   * {BASE_URL}/4_0_1/Library/$package or {BASE_URL}/4_0_1/Library/:id/$package
+   * creates a bundle of the library (specified by parameters) and all dependent libraries
+   * requires parameters id and/or url, but also supports version as supplemental (optional)
+   */
+  async package(args: RequestArgs, { req }: RequestCtx) {
+    logger.info(`${req.method} ${req.path}`);
+
+    const params = gatherParams(req.query, args.resource);
+    const id = args.id || params.id;
+    const url = params.url;
+    const version = params.version;
+
+    if (!id && !url) {
+      throw new BadRequestError('Must provide identifying information via either id or url parameters');
+    }
+
+    // query construction
+    const query: Filter<any> = {};
+    if (id) query.id = id;
+    if (url) query.url = url;
+    if (version) query.version = version;
+    const library = await findResourcesWithQuery<fhir4.Library>(query, 'Library');
+    if (!library || !(library.length > 0)) {
+      throw new ResourceNotFoundError(
+        `No resource found in collection: Library, with ${Object.keys(query)
+          .map(key => `${key}: ${query[key]}`)
+          .join(' and ')}`
+      );
+    }
+    if (library.length > 1) {
+      throw new BadRequestError(
+        `Multiple resources found in collection: Library, with ${Object.keys(query)
+          .map(key => `${key}: ${query[key]}`)
+          .join(' and ')}. /Library/$package operation must specify a single Library`
+      );
+    }
+
+    return createLibraryPackageBundle(library[0]);
   }
 }
--- a/src/services/MeasureService.ts
+++ b/src/services/MeasureService.ts
@@ -57,10 +57,9 @@ export class MeasureService implements Service<fhir4.Measure> {
     const identifier = params.identifier;
 
     if (!id && !url && !identifier) {
-      throw new BadRequestError('Must provide identifying information via either id or url parameters');
+      throw new BadRequestError('Must provide identifying information via either id, url, or identifier parameters');
     }
 
-    // query construction
     const query: Filter<any> = {};
     if (id) query.id = id;
     if (url) query.url = url;
--- a/src/services/MeasureService.ts
+++ b/src/services/MeasureService.ts
@@ -54,8 +54,9 @@ export class MeasureService implements Service<fhir4.Measure> {
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
+    const identifier = params.identifier;
 
-    if (!id && !url) {
+    if (!id && !url && !identifier) {
       throw new BadRequestError('Must provide identifying information via either id or url parameters');
     }
 
@@ -64,7 +65,10 @@ export class MeasureService implements Service<fhir4.Measure> {
     if (id) query.id = id;
     if (url) query.url = url;
     if (version) query.version = version;
-    const measure = await findResourcesWithQuery<fhir4.Measure>(query, 'Measure');
+    if (identifier) query.identifier = identifier;
+
+    const parsedQuery = getMongoQueryFromRequest(query);
+    const measure = await findResourcesWithQuery<fhir4.Measure>(parsedQuery, 'Measure');
     if (!measure || !(measure.length > 0)) {
       throw new ResourceNotFoundError(
         `No resource found in collection: Measure, with ${Object.keys(query)
--- a/src/util/bundleUtils.ts
+++ b/src/util/bundleUtils.ts
@@ -37,7 +37,7 @@ export async function createMeasurePackageBundle(measure: fhir4.Measure): Promis
     const mainLib = libs[0];
 
     const result = await createDepLibraryBundle(mainLib);
-    result.entry?.push({ resource: measure });
+    result.entry?.unshift({ resource: measure });
 
     return result;
   } else {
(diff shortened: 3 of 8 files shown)


p9g:
Many of my comments in https://jira.hl7.org/browse/FHIR-39729 about Measure/$package apply to Library/$package.  Looks like many of the IN parameters are not implemented or tested, which is reasonable for initial implementation.  `identifier `is another IN parameter that can be used to find the specific Library instance, but if using https://github.com/cqframework/ecqm-content-r4-2021/blob/master/bundles/measure/ColorectalCancerScreeningsFHIR/ColorectalCancerScreeningsFHIR-bundle.json for testing, it does not have the identifier on the Library.

test/services/LibraryService.test.ts
```diff
@@ -86,6 +109,95 @@ describe('LibraryService', () => {
...
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('BadRequest');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Must provide identifying information via either id or url parameters'
+          );
+        });
+    });
+
+    it('throws a 404 error when no library matching id and url can be found', async () => {
```
p9g, line 181: See my comment about identifier as another way to find the Library to package, and my HL7 Jira issue for Measure that suggests OperationOutcome error if more than one way of specifying the instance is passed in, but that needs discussion.  What about tests where both id and url is specified but one is invlaid and the other is valid.

src/util/bundleUtils.ts
```diff
@@ -36,23 +36,42 @@ export async function createMeasurePackageBundle(measure: fhir4.Measure): Promis
...
-    const allLibs = Array.from(idMap.values());
-    const result: fhir4.Bundle = { resourceType: 'Bundle', type: 'collection', id: v4() };
-    result.entry = [{ resource: measure }];
-    result.entry.push(
-      ...allLibs.map(r => ({
-        resource: r
-      }))
-    );
+    const result = await createDepLibraryBundle(mainLib);
+    result.entry?.push({ resource: measure });
```
mgramigna, line 40: nitpick, but it probably makes sense to push the measure onto the beginning of the list with [unshift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift) instead 

```suggestion
    result.entry?.unshift({ resource: measure });
```

test/services/LibraryService.test.ts
```diff
@@ -13,17 +20,32 @@ const LIBRARY_WITH_URL: fhir4.Library = {
   url: 'http://example.com'
 };
 
-const LIBRARY: fhir4.Library = {
+const LIBRARY_WITH_NO_DEPS: fhir4.Library = {
```
mgramigna, line 23: This fixture seems to be exactly the same as `LIBRARY`. Is there a reason we can't just replace LIBRARY with this one now?

test/services/LibraryService.test.ts
```diff
@@ -13,17 +20,32 @@ const LIBRARY_WITH_URL: fhir4.Library = {
...
+
+const LIBRARY_WITH_DEPS: fhir4.Library = {
+  resourceType: 'Library',
+  id: 'testLibraryWithDeps',
+  status: 'draft',
+  url: 'http://example.com/testLibraryWithDeps',
   type: { coding: [{ code: 'logic-library' }] },
-  id: 'test',
-  status: 'active'
+  relatedArtifact: [
```
mgramigna, line 37: It may be useful to test one more level of traversal. Like A which depends on B which depends on C and making sure all 3 are included

mgramigna (changes requested):
Looks/works great. Small comments

src/util/bundleUtils.ts
```diff
@@ -36,23 +36,42 @@ export async function createMeasurePackageBundle(measure: fhir4.Measure): Promis
     }
     const mainLib = libs[0];
 
-    const allLibsDups = await getAllDependentLibraries(mainLib as fhir4.Library);
```
sarahmcdougall, line 39: Smart thinking with refactoring this into its own function!

test/services/MeasureService.test.ts
```diff
@@ -137,7 +142,7 @@ describe('MeasureService', () => {
         });
     });
 
-    it('returns a Bundle including the root lib, Measure, and  when root lib has dependencies and id passed through body', async () => {
+    it('returns a Bundle including the root lib, Measure, and when root lib has no dependencies and id passed through body', async () => {
```
sarahmcdougall, line 145:
```suggestion
    it('returns a Bundle including the root lib and Measure when root lib has no dependencies and id passed through body', async () => {
```
I think this is what we want for the test description? Unless something else is supposed to come after the "and" that I'm missing

test/services/MeasureService.test.ts
```diff
@@ -152,6 +157,42 @@ describe('MeasureService', () => {
...
+            expect.arrayContaining([
+              { resource: LIBRARY_WITH_DEPS },
+              { resource: LIBRARY_WITH_NO_DEPS },
+              { resource: MEASURE_WITH_ROOT_LIB_AND_DEPS }
+            ])
+          );
+        });
+    });
+
+    it('returns a Bundle including the root lib, Measure, and when root lib has dependencies and id passed through body', async () => {
```
sarahmcdougall, line 177: Same comment as above.. I think we want to change this to "root Lib and Measure" instead of "root lib, Measure, and"

test/services/MeasureService.test.ts
```diff
@@ -166,7 +207,7 @@ describe('MeasureService', () => {
         });
     });
 
-    it('throws a 404 error when no measure matching id and url cannot be found', async () => {
+    it('throws a 404 error when no measure matching id and url can be found', async () => {
```
sarahmcdougall, line 210: lol good catch on the double negative here

test/services/LibraryService.test.ts
```diff
@@ -86,6 +109,95 @@ describe('LibraryService', () => {
...
+        .get('/4_0_1/Library/testLibraryWithNoDeps/$package')
+        .expect(200)
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Bundle');
+          expect(response.body.entry).toHaveLength(1);
+          expect(response.body.entry).toEqual(expect.arrayContaining([{ resource: LIBRARY_WITH_NO_DEPS }]));
+        });
+    });
+
+    it('returns a Bundle including the Library, and when the Library has no dependencies and id passed through body', async () => {
```
sarahmcdougall, line 125:
```suggestion
    it('returns a Bundle including the Library when the Library has no dependencies and id passed through body', async () => {
```

test/services/LibraryService.test.ts
```diff
@@ -86,6 +109,95 @@ describe('LibraryService', () => {
...
+        .set('content-type', 'application/fhir+json')
+        .expect(200)
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Bundle');
+          expect(response.body.entry).toHaveLength(1);
+          expect(response.body.entry).toEqual(expect.arrayContaining([{ resource: LIBRARY_WITH_NO_DEPS }]));
+        });
+    });
+
+    it('returns a Bundle including the Library, and when the Library has dependencies and id passed through args', async () => {
```
sarahmcdougall, line 138:
```suggestion
    it('returns a Bundle including the Library and its dependencies when the Library has dependencies and id passed through args', async () => {
```

test/services/LibraryService.test.ts
```diff
@@ -86,6 +109,95 @@ describe('LibraryService', () => {
...
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Bundle');
+          expect(response.body.entry).toHaveLength(2);
+          expect(response.body.entry).toEqual(
+            expect.arrayContaining([{ resource: LIBRARY_WITH_DEPS }, { resource: LIBRARY_WITH_NO_DEPS }])
+          );
+        });
+    });
+
+    it('returns a Bundle including the Library, and when the Library has dependencies and id passed through body', async () => {
```
sarahmcdougall, line 152:
```suggestion
    it('returns a Bundle including the Library and its dependencies when the Library has dependencies and id passed through body', async () => {
```

test/services/LibraryService.test.ts
sarahmcdougall, line 181: I don't think we have a Measure/$package test with both id and url specified but one is invalid and the other is valid, so if we add a test for that here then we should also do the same for Measure/$package

sarahmcdougall (changes requested):
Great work with this! Very clear to see what's going on here, and thanks for including detailed testing guidance :) Just a few small suggestions/comments

test/services/LibraryService.test.ts
elsaperelli: Good suggestion!

src/util/bundleUtils.ts
elsaperelli: Matt's idea!

src/util/bundleUtils.ts
```diff
@@ -36,23 +36,42 @@ export async function createMeasurePackageBundle(measure: fhir4.Measure): Promis
...
+  const result = await createDepLibraryBundle(library);
+
+  return result;
+}
+
+/**
+ * Takes in the main library from either Measure/$package or Library/$package
+ * and returns a bundle of all the dependent libraries
+ */
+export async function createDepLibraryBundle(mainLib: fhir4.FhirResource): Promise<fhir4.Bundle<fhir4.FhirResource>> {
```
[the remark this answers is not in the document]
mgramigna: good call Sarah. We shouldn't even need to cast. We should just be able to use the generic on `findResourcesWithQuery`:

```ts
const libs = await findResourcesWithQuery<fhir4.Library>(mainLibQuery, 'Library');
```

and then update the signature of `createDepLibraryBundle` to take a `fhir4.Library`

test/services/LibraryService.test.ts
elsaperelli: Sorry this is late! I am working on adding identifier, but I just added your testing suggestion. Thank you!

test/services/LibraryService.test.ts
```diff
@@ -13,27 +13,56 @@ const LIBRARY_WITH_URL: fhir4.Library = {
...
-  status: 'active'
+  relatedArtifact: [
+    {
+      type: 'depends-on',
+      resource: 'http://example.com/testLibrary'
+    }
+  ]
+};
+
+const LIBRARY_WITH_DEPS: fhir4.Library = {
```
sarahmcdougall, line 38: This is minor but maybe we can consider renaming `LIBRARY_WITH_DEPS` to `LIBRARY_WITH_NESTED_DEPS` and rename `LIBRARY_WITH_DEPS_2` to `LIBRARY_WITH_DEPS`. This renaming makes it easier for the developer to quickly see why we have multiple library fixtures with dependencies

elsaperelli: Good suggestion!

test/services/LibraryService.test.ts
```diff
@@ -86,6 +147,178 @@ describe('LibraryService', () => {
...
+          expect(response.body.entry).toEqual(expect.arrayContaining([{ resource: LIBRARY_WITH_IDENTIFIER_VALUE }]));
+        });
+    });
+
+    it('returns a Bundle including just the Library when the Library has no dependencies and identifier with just identifier.system passed through body', async () => {
+      await supertest(server.app)
+        .post('/4_0_1/Library/$package')
+        .send({
+          resourceType: 'Parameters',
+          parameter: [{ name: 'identifier', valueString: 'http://example.com/libraryWithIdentifierSystem|' }]
```
p9g, line 234: What if there are multiple Libraries with the same identifier system but different values?  Unless you are going to package them all, probably should expect 400.  IG should be clearer about how to handle such cases.

elsaperelli: I just added a test for this in my latest [commit](https://github.com/projecttacoma/measure-repository-service/pull/6/commits/b4d1898f62605f05632ecc2eace28683e5e56765). I hope this is what you had in mind!

src/services/LibraryService.ts
```diff
@@ -39,4 +40,50 @@ export class LibraryService implements Service<fhir4.Library> {
...
+    logger.info(`${req.method} ${req.path}`);
+
+    const params = gatherParams(req.query, args.resource);
+    const id = args.id || params.id;
+    const url = params.url;
+    const version = params.version;
+    const identifier = params.identifier;
+
+    if (!id && !url && !identifier) {
+      throw new BadRequestError('Must provide identifying information via either id or url parameters');
```
mgramigna, line 60: Or identifier?

src/services/LibraryService.ts
```diff
@@ -39,4 +40,50 @@ export class LibraryService implements Service<fhir4.Library> {
...
+    }
+
+    // query construction
```
mgramigna, line 63: code is self documenting ;) 
```suggestion
```

src/services/MeasureService.ts
```diff
@@ -54,8 +54,9 @@ export class MeasureService implements Service<fhir4.Measure> {
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
+    const identifier = params.identifier;
 
-    if (!id && !url) {
+    if (!id && !url && !identifier) {
```
mgramigna, line 59: Or identifier in error message

src/util/queryUtils.ts
```diff
@@ -18,7 +18,7 @@ export function getMongoQueryFromRequest(query: RequestQuery): Filter<any> {
       // For string arguments in query they may match just the start of a string and are case insensitive
       mf[key] = { $regex: `^${query[key]}`, $options: 'i' };
     } else if (key === 'identifier') {
-      // Identifier can check against the identifier.code, identifier.value, or both on a resource
+      // Identifier can check against the identifier.system, identifier.value, or both on a resource
```
mgramigna, line 21: good catch

mgramigna (changes requested):
Looks great, small comments

mgramigna (approved):

sarahmcdougall (approved):

PR #7, opened 2023-01-06
Update OperationOutcome issue codes

sarahmcdougall:
# Summary
Updates the OperationOutcome `issue.code` for all thrown OperationOutcomes to be of type `IssueType`.

## New behavior
When an error is thrown from the server, the resulting OperationOutcome will have an `issue.code` that is a valid `IssueType` - see the [code system](https://www.hl7.org/fhir/codesystem-issue-type.html) for more information on what codes are considered valid.

## Code changes
* Updates to `errorUtils.ts` to change the `issue.code`s for the custom error classes. `ResourceNotFoundError` and `NotImplementedError` are pretty straightforward. Since `BadRequestError` is pretty broad, I decided to have the error default to `issue.code` ‘invalid’ (which is the “parent” of invalid codes). Throughout the code, if a more specific code made sense (ex. ‘required’ instead of ‘invalid’ if the request is missing a required element), I replaced ‘invalid’ with that code. Open to suggestions on this since these changes were made based on my own intuition — if it makes sense to just have all `BadRequestError`s be ‘invalid’ then that would also work.
* Updates to corresponding unit tests, comments, etc. 

# Testing guidance
* Run unit tests, as all tests containing an OperationOutcome check the value of the `issue.code`
* Try testing in the server - do something that would throw an OperationOutcome (ex. Send a GET request to a library that does not exist on the server) and check that the code is of type `IssueType`

### DIFF
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -57,7 +57,10 @@ export class LibraryService implements Service<fhir4.Library> {
     const identifier = params.identifier;
 
     if (!id && !url && !identifier) {
-      throw new BadRequestError('Must provide identifying information via either id, url, or identifier parameters');
+      throw new BadRequestError(
+        'Must provide identifying information via either id, url, or identifier parameters',
+        'required'
+      );
     }
 
     const query: Filter<any> = {};
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -1,4 +1,4 @@
-import { loggers, RequestArgs, RequestCtx } from '@projecttacoma/node-fhir-server-core';
+import { loggers, RequestArgs, RequestCtx, constants } from '@projecttacoma/node-fhir-server-core';
 import { Filter } from 'mongodb';
 import { findResourceById, findResourcesWithQuery } from '../db/dbOperations';
 import { Service } from '../types/service';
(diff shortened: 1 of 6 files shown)


mgramigna (changes requested):
Just noticed. All the codes are defined [here](https://github.com/projecttacoma/measure-repository-service/blob/main/src/types/node-fhir-server-core.d.ts). It's not pretty, but we can probably just do 

```ts
import { constants } from '@projecttacoma/node-fhir-server-core';
// ...
constants.ISSUE.CODE.INVALID;

sarahmcdougall:
> Just noticed. All the codes are defined [here](https://github.com/projecttacoma/measure-repository-service/blob/main/src/types/node-fhir-server-core.d.ts).

Good point - I noticed that all the codes are defined here when writing up the test server task in Jira but I didn't think to import it into this project since we don't throw many errors. But it's probably better to be safe and use these constants to avoid typos, so I'll try importing and see if it works out 👍

mgramigna:
> > Just noticed. All the codes are defined [here](https://github.com/projecttacoma/measure-repository-service/blob/main/src/types/node-fhir-server-core.d.ts).
> 
> Good point - I noticed that all the codes are defined here when writing up the test server task in Jira but I didn't think to import it into this project since we don't throw many errors. But it's probably better to be safe and use these constants to avoid typos, so I'll try importing and see if it works out 👍

Yeah. There are more typescripty ways to do it that maybe some motivated person will implement at some point, but for now let's just use the contants

mgramigna (approved):

elsaperelli (approved):
Lgtm! Also, I agree with you about the defaulting it to invalid but specifying more if the information is available. Good thinking!

PR #9, opened 2023-01-09
Data requirements

lmd59:
# Summary
Creates the `/4_0_1/Measure/$data-requirements` endpoint which first uses the $package function to create a measure bundle and then uses fqm-execution `Calculator.calculateDataRequirements` to get the data requirements Library for the bundle, as outlined (sort of) in the [docs](http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Measure-data-requirements)

Note: There are two outstanding questions about these docs, which have to do with the return type (Library vs. Bundle) and whether the measurement period should be required. Current behavior defaults the measurement period to 2022 and assumes a `Library` resource should be returned. TODOs in code outline potential future changes may be required.

## New behavior
 - 4 new endpoints:
 - `GET 4_0_1/Measure/$data-requirements`
 - `GET 4_0_1/Measure/<id>/$data-requirements`
 - `POST 4_0_1/Measure/$data-requirements`
 - `POST 4_0_1/Measure/<id>/$data-requirements`

## Code changes
 - Added new endpoints in `serverConfig`
 - New function in `MeasureService`
 - `README` update
 - Added testing

# Testing guidance
 - `npm run check`
 - `npm run db:reset`
 - ```npm run db:loadBundle "ecqm-content-r4-2021/bundles/measure/ColorectalCancerScreeningsFHIR/ColorectalCancerScreeningsFHIR-bundle.json"```
 - `npm start`
 - Use various combinations of endpoints with the $data-requirements operation. Example: GET to http://localhost:3000/4_0_1/Measure/ColorectalCancerScreeningsFHIR/$data-requirements?periodStart=2022-01-01&periodEnd=2022-12-31
 - The $package [Insomnia Requests](https://github.com/projecttacoma/measure-repository-service/files/10281840/Measure-.package-testing.json.zip) can be adapted by replacing `$package` with `$data-requirements`
 - Also play with different ways of including the measurement period parameters
 - The returned Library should include data requirements that make sense for the colorectal cancer measure (I think 45 of them should be expected)


src/services/MeasureService.ts
```diff
@@ -88,4 +89,34 @@ export class MeasureService implements Service<fhir4.Measure> {
 
     return createMeasurePackageBundle(measure[0]);
   }
+
+  /**
+   * result of sending a POST or GET request to:
+   * {BASE_URL}/4_0_1/Measure/$data-requirements or {BASE_URL}/4_0_1/Measure/:id/$data-requirements
+   * creates a Library with all data requirements for the specified measure
+   * requires parameters id and/or url and/or identifier, but also supports version as supplemental (optional)
```
elsaperelli, line 97: Should we add something about periodStart and periodEnd in this? I know this might change

test/services/MeasureService.test.ts
```diff
@@ -345,6 +346,74 @@ describe('MeasureService', () => {
...
+        .post('/4_0_1/Measure/testWithRootLibAndDeps/$data-requirements')
+        .set('content-type', 'application/fhir+json')
+        .expect(200)
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Library');
+          expect(response.body.dataRequirement).toHaveLength(0);
+        });
+    });
+
+    it('returns 200 with passed period parameters', async () => {
```
elsaperelli, line 386: If no period parameters are passed in, the code defaults the periodStart to 2022-01-1 and periodEnd to 2022-12-31. In that case, I think this test might be more beneficial if the passed in periodStart and periodEnd were different (ex. 2018-01-01 and 2018-12-31) if that makes sense!

test/services/MeasureService.test.ts
```diff
@@ -345,6 +346,74 @@ describe('MeasureService', () => {
...
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Library');
+          expect(response.body.dataRequirement).toHaveLength(0);
+        });
+    });
+
+    it('returns 200 and a Library for a request with id in the url', async () => {
+      await supertest(server.app)
+        .post('/4_0_1/Measure/testWithRootLibAndDeps/$data-requirements')
+        .set('content-type', 'application/fhir+json')
```
elsaperelli, line 378:
```suggestion
        .get('/4_0_1/Measure/testWithRootLibAndDeps/$data-requirements')
```
I think we can do this since nothing is being passed through the body (similar to test on line 163 for Measure/$package).

test/services/MeasureService.test.ts
```diff
@@ -345,6 +346,74 @@ describe('MeasureService', () => {
...
+        .then(response => {
+          expect(response.body.resourceType).toEqual('Library');
+          expect(response.body.dataRequirement).toHaveLength(0);
+        });
+    });
+
+    it('returns 200 with query params', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Measure/$data-requirements?id=testWithRootLibAndDeps&periodStart=2022-01-01&periodEnd=2022-12-31')
+        .set('content-type', 'application/fhir+json')
```
elsaperelli, line 408:
```suggestion
```
I don't think we need this since it's a get request.

test/services/MeasureService.test.ts
elsaperelli: Upon further inspection, I don't think changing this will change anything. Is there any other way to test the periodStart and periodEnd parameters though? Or does fqm-execution mostly change that testing.

elsaperelli (changes requested):
Nice! Just a few comments/suggestions/questions :)

src/services/MeasureService.ts
```diff
@@ -88,4 +89,34 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+   */
+  async dataRequirements(args: RequestArgs, { req }: RequestCtx) {
+    logger.info(`${req.method} ${req.path}`);
+    logger.info('Using package to create measure bundle');
+    const measureBundle = await this.package(args, { req: req });
```
sarahmcdougall, line 102: Adding a logger statement about using the package operation is a good idea 👍 I did notice that when passing `req` into `this.package`, the `package` logger statement says the `req.path` is data requirements instead of package:

```
{"level":"info","message":"GET /4_0_1/Measure/ColorectalCancerScreeningsFHIR/$data-requirements"}
{"level":"info","message":"Using package to create measure bundle"}
{"level":"info","message":"GET /4_0_1/Measure/ColorectalCancerScreeningsFHIR/$data-requirements"}
{"level":"info","message":"Assembling collection bundle from Measure ColorectalCancerScreeningsFHIR"}
```
This makes it look like data requirements is being called twice. Maybe we should pass in an updated path to `this.package` so that its logger is accurate

test/services/MeasureService.test.ts
```diff
@@ -345,6 +346,74 @@ describe('MeasureService', () => {
...
+      }
+    });
+
+    it('returns 200 and a Library for a simple measure with dependencies and no period params', async () => {
+      await supertest(server.app)
+        .post('/4_0_1/Measure/$data-requirements')
+        .send({ resourceType: 'Parameters', parameter: [{ name: 'id', valueString: 'testWithRootLibAndDeps' }] })
+        .set('content-type', 'application/fhir+json')
+        .expect(200)
+        .then(response => {
```
sarahmcdougall, line 369: According to the spec: "The result of the requirements operation is a _module-definition_ Library that returns the computed effective requirements of the artifact." Do you think it's worth adding a check in these unit tests that the library-type of the returned FHIR Library is "module-definition"?

sarahmcdougall (changes requested):
Good work with communicating the discrepancies in the spec for data requirements - I think this PR makes a good call with assuming a Library is returned and defaulting the measurement period. 

Just have two in-line comments that may/may not warrant changes but overall the functionality and code are both good 👍

README.md
```diff
@@ -57,6 +57,15 @@ The Measure Repository Service server supports `Library` and `Measure` resource
 - url
 - version
 
+### Data Requirements
```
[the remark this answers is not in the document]
lmd59: Sure! Good call.

test/services/MeasureService.test.ts
lmd59: Y'know. I didn't think there was, but you made me think through it a bit more, and there's totally a way to check the dates using the jest spy. I'll make that change!

src/services/MeasureService.ts
lmd59: Hmm- things seem to go a bit sideways when I start futzing with the path... I can experiment with it more, but did you have any more thoughts on implementation you imagine?

test/services/MeasureService.test.ts
lmd59: I can add this, but looking at it, it's really just checking the values on the mocked return item just above, created by the spy. Maybe it's still good to have from future-proof perspective if we ever go in a different testing direction.

README.md
```diff
@@ -59,13 +74,18 @@ The Measure Repository Service server supports `Library` and `Measure` resource
...
 
-The Measure Repository Service server supports the `Measure` `$data-requirements` operation with the SHALL parameters specified in the publishable measure repository section of the [HL7 Measure Repository Docs](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#publishable-measure-repository). These parameters are:
+The Measure Repository Service server supports the `Measure` `$data-requirements` operation with the SHALL parameters specified in the publishable measure repository section of the [HL7 Measure Repository Docs](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#publishable-measure-repository). SHALL parameters are:
 
 - id
 - url
 - version
 - identifier
 
+Supported optional optional parameters are:
```
elsaperelli, line 84: Optional optional? :P

test/services/MeasureService.test.ts
```diff
@@ -390,26 +409,46 @@ describe('MeasureService', () => {
...
+          );
         });
     });
 
     it('returns 200 with query params', async () => {
       await supertest(server.app)
-        .get('/4_0_1/Measure/$data-requirements?id=testWithRootLibAndDeps&periodStart=2022-01-01&periodEnd=2022-12-31')
-        .set('content-type', 'application/fhir+json')
+        .get('/4_0_1/Measure/$data-requirements')
+        .query({ id: 'testWithRootLibAndDeps', periodStart: '2021-01-01', periodEnd: '2021-12-31' })
```
elsaperelli, line 437: Right now, we test periodStart and periodEnd when they are passed through
1. Parameter
2. Query
Would it be beneficial to add one more that passes the periodStart and periodEnd through the url like this test used to? I think this was a good change here but maybe we can have both? Or is that redundant? If we do add it, probably would be good to use 2021 like the others :)

elsaperelli (changes requested):

src/services/MeasureService.ts
sarahmcdougall: If it's too complicated to fix then I'd say we can leave this since it's such a small thing. I tried this out and it seemed to work alright:

``` javascript
const packagePath = req.path.replace(/data-requirements/i, 'package');
const measureBundle = await this.package(args, { req: { ...req, path: packagePath } });
```
which returned `"GET /4_0_1/Measure/ColorectalCancerScreeningsFHIR/$package"`

lmd59: oh interesting, I tried to directly futz with req.path and got that it was readonly, but yeah, this seems to work. Thank you for digging further!

test/services/MeasureService.test.ts
lmd59: I believe that this replacement is equivalent to putting the query parameters directly into the url. I made this change to standardize with other tests in this file that use `.query` instead.

src/services/MeasureService.ts
lmd59: Oh- also I like your replacement approach better than mine which was just `req.path.replace('data-requirements', 'package');` :-)

README.md
lmd59: lol oops!

test/services/MeasureService.test.ts
elsaperelli: Sounds good!

elsaperelli (changes requested):
I think `ColorectalCancerScreeningsFHIR-bundle.json` and `.DS_Store` were accidentally added- other than that I am ready to approve!

sarahmcdougall (commented):
I second Elsa's observation - looks like `.DS_Store` and the Colorectal Cancer Screening bundle sneaked in (maybe a result of the rebase). I think both can be safely deleted, and it probably doesn't hurt to add `.DS_Store` to the project's `.gitignore`

Also, putting this here for reference in case someone looks back on the PR later: the period start/end params should not have default values. Matt mentioned in slack that he updated the measure repo service to not default period start/end in data requirements. Future work in fqm-execution will make period start/end optional for data requirements.

src/services/MeasureService.ts
```diff
@@ -88,4 +89,35 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+    const { results } = await Calculator.calculateDataRequirements(measureBundle, {
+      measurementPeriodStart: start,
+      measurementPeriodEnd: end
+    });
+    logger.info('Successfully generated $data-requirements report');
+    return results;
+
+    // TODO: Clarify return type from documentation (https://hl7.org/fhir/us/cqfmeasures/STU3/OperationDefinition-Measure-data-requirements.html)
+    // Documentation describes as a library, but the specified type is "Bundle".
+    // For now, this code returns a Library (similar to the Measure operation)
```
sarahmcdougall, line 121: I just realized that Matt said in Slack that the return type was confirmed to be Library (see Jira ticket [here](https://jira.hl7.org/browse/FHIR-40226) with the correction). Do we think it's safe to remove this TODO?

elsaperelli: I agree, Sarah. I think we can remove that TODO.

lmd59:
> I think `ColorectalCancerScreeningsFHIR-bundle.json` and `.DS_Store` were accidentally added- other than that I am ready to approve!

oy vey! *head smack* Sorry about that.

src/services/MeasureService.ts
```diff
@@ -88,4 +89,31 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+   * requires parameters id and/or url and/or identifier, but also supports version as supplemental (optional)
+   */
+  async dataRequirements(args: RequestArgs, { req }: RequestCtx) {
+    logger.info(`${req.method} ${req.path}`);
+    logger.info('Using package to create measure bundle');
+    const path = req.path.replace(/data-requirements/i, 'package');
+    const measureBundle = await this.package(args, { req: { ...req, path: path } });
+
+    // TODO: Clarify should period start/end be required or have intelligent defaults in fqm-execution to measure effectivePeriod? (fqm-execution defaults to 2019)
+    // short term: default to 2022 if these aren't passed through
```
elsaperelli, line 106: My last question (promise!) is whether we should change this TODO based on what Matt learned at the connectathon.

elsaperelli (approved):
Lgtm! Just one last question but if you don't think that needs to be changed then it's good to go from me :)

src/services/MeasureService.ts
sarahmcdougall: Matt's [connectathon branch](https://github.com/projecttacoma/measure-repository-service/commit/615f33a5fa635b5301ebeb47436c91138557ab29) makes the update and removes this TODO, so I think it's fine to leave it here since the actual changes will come through in his PR (meaning this will be in the main branch for a v short period of time)

sarahmcdougall (approved):

PR #12, opened 2023-01-20
Add modelinfo to package operations

elsaperelli:
# Summary
After the January '23 connectathon, Matt realized that our `$package` operations should include the `FHIR-ModelInfo` dependency. `include-dependencies` means every dependency, not just dependent library resources. The reason why we didn't include this before is because the measure bundles in `ecqm-content-r4-2021` actually didn't have the `FHIR-ModelInfo` library. 

## New behavior
Since we now want to include this library but it is not contained in the `ecqm-content-r4-2021` bundles, we decided the best approach is to always include the `FHIR-ModelInfo` library in the measure-repository-service. In order to do this, we added the raw json for this library to a new `fixtures` directory in the `scripts` directory and then add it to the Library collection when the database is setup. Let me know if this approach doesn't make sense or if you have any ideas on how to improve this.

## Code changes
- `scripts/dbSetup.ts` - new `insertFHIRModelInfoLibrary()` function that gets called when the database is either setup or reset. I was going to put this code directly in the `createCollections()` function, but I thought it would be more organized to have it in its own function and then add `.then` statements for it after `createCollections()` is called. This approach can be changed based on what makes the most sense.
- `scripts/fixtures/Library-FHIR-ModelInfo.json` - this was found [here](http://fhir.org/guides/cqf/common/Library-FHIR-ModelInfo.json.html). The structure of the files can also change to what makes the most sense.
- `src/utils/bundleUtils.ts` - removed line of code that previously excluded the ModelInfo

# Testing guidance
Okay there is a lot going on with testing so bear with me. We want to test 2 things: $package operations (both Measure and Library) now include `FHIR-ModelInfo` AND the resulting bundle is executable in `deqm-test-server`. Let me know if there is anything additional we should test regarding this.

**TEST 1: $package operations**
1. Below I have a .zip of all the necessary Insomnia tests. Unzip the file and upload it into Insomnia (bottom left corner "Preferences" -> Data -> Import Data -> From File -> you got it 👍 ) 
2. Since we add this library in the script, we have to reset the database.
3. `npm run db:reset`
4. It should say that it deleted and then created 3 collections followed by `Inserting Library/FHIR-ModelInfo into database`
5. `git clone https://github.com/cqframework/ecqm-content-r4-2021.git`
6. `npm run db:loadBundle "ecqm-content-r4-2021/bundles/measure/ColorectalCancerScreeningsFHIR/ColorectalCancerScreeningsFHIR-bundle.json"`
7. `npm start`
8. Open the `FHIRModelInfo Testing` collection in Insomnia.
9. The first two `Measure/$package` tests should result in a `202 OK` as well as return bundles with 11 entries. This is similar to Nat's original Measure/$package [PR](https://github.com/projecttacoma/measure-repository-service/pull/5) where he says the bundle should include the Measure itself as well as 9 Library resources but now we have 10 Library resources because `FHIR-ModelInfo` is also included. Make sure the rest are the same.
10. The next two `Library/$package` tests should result in the same thing except for the Measure itself. So now 10 Library resources (including FHIR-ModelInfo).
11. Export a bundle from a Measure/$package test (either one) to a .json file to be used later.

**TEST 2: executable?**
1. Use that bundle.json file now
2. Pull up `deqm-test-server`
3. `npm run db:delete`
4. `npm run db:setup`
5. `npm run upload-bundles <path-to-folder-with-that-measure-package-output-json-file.json> "name-of-that-file"` (example: `npm run upload-bundles ./../../../Downloads "measure-package-output.json`)
6. It should say that 11 resources were uploaded from 1 Bundle files.
7. Go back to Insomnia to the `deqm-test-server` collection and run through the included Insomnia requests. I just included Library search so you can make sure that `FHIR-ModelInfo` is there. 
8. The goal of this is to "Make sure the resulting measure is executable in deqm-test-server" so if you think of any other necessary testing for that goal, let me know!

Insomnia files: [FhirModelInfoTesting.json.zip](https://github.com/projecttacoma/measure-repository-service/files/10468910/FhirModelInfoTesting.json.zip)


mgramigna (approved):

natjoe4 (approved):
Nice job. Clean code and testing works well :)

PR #14, opened 2023-01-30
Add custom capability statement

sarahmcdougall:
# Summary
Adds a custom capability statement of kind “instance” that covers the current capabilities of the measure repository service. It is accessible via the `/metadata` endpoint.

## New behavior
Creates a custom capability statement that reflects the current state of the measure repository service, specifically in regard to the requirements for a Shareable Measure Repository and Publishable Measure Repository. Note the use of the kind “instance,” which requires an implementation to be present and describes its capabilities. Whereas the [Publishable M.R.’s capability statement](https://build.fhir.org/ig/HL7/cqf-measures/CapabilityStatement-publishable-measure-repository.json.html) imports the capability statement for the Shareable M.R., this instance capability statement does not (this is something we can change in the future if the single capability statement gets too unruly, but for now I think it’s small enough to include everything in a single cap. statement.)

This capability statement currently only includes the SHALL operations and search params that we support. As we implement SHOULD operations and search params, we will need to update the capability statement accordingly and amend its date (the date represents the date last updated).

To access the capability statement, send a GET request to `http://localhost:3000/4_0_1/metadata`.

## Code changes
* Updated the `serverConfig` to create a new custom capability statement with the `statementGenerator` that was recently added to `node-fhir-server-core.d.ts`
* Added a JSON file containing the definition of RESTful capabilities of the server

Some design decisions (that I’m open to feedback on):
* I removed the `RequestArgs` input from the `statementGenerator` and instead defaulted to version 4_0_1. In the future I imagine we want to support multiple FHIR versions, but for now I figure it is fine to default to 4_0_1 since we default the Measure/Library profiles to only support version 4_0_1
* Node-fhir-server-core does create a default `rest` array based on the defined profiles and operations in the`serverConfig`; however, I found it difficult to really use any of it, which is why I created a separate JSON containing the desired contents of the `rest` array. If you send a GET request to `http://localhost:3000/4_0_1/metadata` from the main branch, you can see the capability statement that gets automatically generated. The resource types contain search params that the server does not support, and there are duplicate operations reflected (due to how we define them in the `serverConfig`). Rather than try to take bits and pieces from this, I think it would be more clear to the developer if we stashed the `rest` contents in a separate JSON.
* The default cap. statement that gets generated includes the `operation` array at the `rest` level. However, the operations can instead be defined for each resource type. According to the [spec](https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.operation), “CapabilityStatement.rest.operation is for operations invoked at the system level, or for operations that are supported across multiple resource types.” When defined on the resource, the [spec](https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.resource.operation) says “If an operation that is listed in multiple CapabilityStatement.rest.resource.operation (e.g. for different resource types), then clients should understand the the operation is only supported on the specified resource types.” From this, I decided to put the `package` and `data-requirements` operations under Library and Measure resources rather than have them inferred as system-level operations.
* I included SHALL extensions (they are used in the requirements cap. statements for Shareable/Publishable) but we can remove them if they aren’t needed. I figure they might be useful for future cap. statement tests in the test kit.

# Testing guidance
* Send GET request to the endpoint (`http://localhost:3000/4_0_1/metadata`)
* Inspect the contents to ensure that they align with the spec
    * all required fields are present
    * fields like “instantiates” and “format” are correct
    * The correct interactions, search params, and operations are present for each resource type
    * Look at the spec and see if there are any optional fields not reflected in the cap. statement that we should include

### DIFF
--- a/src/config/capabilityStatementResources.json
+++ b/src/config/capabilityStatementResources.json
@@ -54,7 +54,7 @@
           ],
           "name" : "version",
           "definition" : "http://hl7.org/fhir/SearchParameter/Library-version",
-          "type" : "token"
+          "type" : "string"
         },
         {
           "extension" : [
@@ -120,14 +120,6 @@
           }],
           "name" : "package",
           "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-package"
-        },
-        {
-          "extension" : [{
-            "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
-            "valueCode" : "SHALL"
-          }],
-          "name" : "data-requirements",
-          "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-data-requirements"
         }
       ]
     },
@@ -186,7 +178,7 @@
           ],
           "name" : "version",
           "definition" : "http://hl7.org/fhir/SearchParameter/Measure-version",
-          "type" : "token"
+          "type" : "string"
         },
         {
           "extension" : [
--- a/src/config/capabilityStatementResources.json
+++ b/src/config/capabilityStatementResources.json
@@ -1,0 +1,267 @@
+{
+  "mode": "server",
+  "resource": [
+    {
+      "type" : "Library",
+      "profile" : "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/library-cqfm",
+      "supportedProfile" : [
+        "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/computable-library-cqfm",
+        "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/publishable-library-cqfm",
+        "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/executable-library-cqfm",
+        "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/modelinfo-library-cqfm",
+        "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/quality-program-cqfm"
+      ],
+      "interaction" : [
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "code" : "read",
+          "documentation" : "Read allows clients to get the definitions and details of repository libraries"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "code" : "search-type",
+          "documentation" : "Search allows clients to search for repository libraries"
+        }
+      ],
+      "searchParam" : [
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "url",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-url",
+          "type" : "uri"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "version",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-version",
+          "type" : "token"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "identifier",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-identifier",
+          "type" : "token"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "name",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-name",
+          "type" : "string"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "title",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-title",
+          "type" : "string"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "description",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-description",
+          "type" : "string"
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "status",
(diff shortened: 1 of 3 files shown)


src/config/serverConfig.ts
```diff
@@ -1,5 +1,29 @@
...
+  const base_version = constants.VERSIONS['4_0_1'];
+  const CapabilityStatement = resolveSchema(base_version, 'CapabilityStatement');
+
+  return new CapabilityStatement({
+    id: 'MeasureRepositoryServiceCapabilityStatement',
+    name: 'MeasureRepositoryService',
+    title: 'FHIR Measure Repository Service Capability Statement',
+    status: 'active',
+    // date last modified
+    date: '2023-01-27',
```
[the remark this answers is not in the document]
elsaperelli: Interesting. I did see though that it is formatted like this in the Publishable Measure Repository service capability statement [here](https://build.fhir.org/ig/HL7/cqf-measures/CapabilityStatement-publishable-measure-repository.json.html). So this probably is correct.

src/config/capabilityStatementResources.json
```diff
@@ -0,0 +1,267 @@
...
+        },
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "version",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Library-version",
```
natjoe4, line 56: The links for these definitions don't lead to a valid page for me. Is that something we need to address?

src/config/capabilityStatementResources.json
```diff
@@ -0,0 +1,267 @@
...
+          "type" : "token"
```
natjoe4, line 57: I believe that "version" is of type string. From the docs, it looks like token search params are typically used for parameters of code or identifier type. Since version is just of type string, I think this should just be string also?

src/config/capabilityStatementResources.json
```diff
@@ -0,0 +1,267 @@
...
+        {
+          "extension" : [
+            {
+              "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+              "valueCode" : "SHALL"
+            }
+          ],
+          "name" : "version",
+          "definition" : "http://hl7.org/fhir/SearchParameter/Measure-version",
+          "type" : "token"
```
natjoe4, line 189: I believe that "version" is of type string. From the docs, it looks like token search params are typically used for parameters of code or identifier type. Since version is just of type string, I think this should just be string also?

src/config/serverConfig.ts
```diff
@@ -1,5 +1,29 @@
...
+    publisher: 'The MITRE Corporation',
+    instantiates: [
+      'http://hl7.org/fhir/us/cqfmeasures/CapabilityStatement/shareable-measure-repository',
+      'http://hl7.org/fhir/us/cqfmeasures/CapabilityStatement/publishable-measure-repository'
```
natjoe4, line 19: publishable-measure-repository capability statement looks like it supports read and search for MeasureReport objects, which the measure repository service doesn't do yet. Should we hold off on adding this until we add read and search capability for MeasureReports?

elsaperelli (changes requested):
Just a few comments/questions! Nice job, I went through the capability statement spec and it looks like you included everything that's needed!

src/config/capabilityStatementResources.json
elsaperelli: Oh wow nice catch haha they also don't work for me. Not sure what it should be instead though?

natjoe4 (changes requested):
Pulling the capability statement works great and looks good! I have a few comments which are just as much questions as they are suggestions. 

One overarching point is that in the CapabilityStatement spec it states "A Capability Statement SHALL have at least one of description, software, or implementation element." I don't believe any of those are found in this capability statement. From what I can tell, since this is of kind 'instance' we need to add `implementation` and can optionally add `software` and `description` although for a first pass, I think it will be fine to leave those out.

src/config/capabilityStatementResources.json
elsaperelli: It looks like the type is string in the Capability Statement docs but then in the reference implementation its of type token ❓

src/config/capabilityStatementResources.json
sarahmcdougall: mattsignal

src/config/capabilityStatementResources.json
sarahmcdougall: Oops yes I agree with you Nat that this should be of type "string" so I'll make that change. The Shareable MR capability statement lists it as a token which seems to be incorrect. Thanks for doing that cross-checking!

src/config/serverConfig.ts
sarahmcdougall: My understanding of the publishable MR is that everything MeasureReport-related is optional since they all appear as `MAY` (ex. "MAY support retrieval of test cases by server-specific id through the MeasureReport/read interaction"), which would mean that we do instantiate the publishable MR

src/config/serverConfig.ts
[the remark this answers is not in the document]
sarahmcdougall: The `date` field in the capability statement is of type `dateTime` (more info [here](https://www.hl7.org/fhir/datatypes.html#dateTime)), and so I'm pretty sure both formats are technically correct. (I just did YYYY-MM-DD since I figured this is as precise as we need it to be.) I'll try the non-hyphenated format and see if it adds the hours/minutes/seconds automatically, but if it doesn't work then I feel like it's not a big deal and we can just use the hyphenated format for now 👍

src/config/capabilityStatementResources.json
```diff
@@ -0,0 +1,267 @@
...
+          "name" : "package",
+          "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-package"
+        },
+        {
+          "extension" : [{
+            "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+            "valueCode" : "SHALL"
+          }],
+          "name" : "data-requirements",
+          "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-data-requirements"
```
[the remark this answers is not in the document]
sarahmcdougall: Good catch, I totally thought we already implemented this lol. I'll add a reminder to the corresponding Jira ticket to update the compatibility statement as part of the task

sarahmcdougall:
> From what I can tell, since this is of kind 'instance' we need to add `implementation` and can optionally add `software` and `description` although for a first pass, I think it will be fine to leave those out.

I added `implementation` to the capability statement with a populated `description`. We can optionally add a `url` to `implementation` but since this isn't hosted anywhere, I figured we shouldn't include it, unless we want to do something like `example.com` as a placeholder or `localhost`

src/config/capabilityStatementResources.json
elsaperelli: Ah I actually found something interesting completely by accident looking through the spec for my other task so I could be wrong. But in the FHIR search docs, it says "A token type is a parameter that provides a close to exact match search on a string of characters, potentially scoped by a URI" which I believe is what we want here since search for version is an exact match but search for other string types (title, name, description) are not exact matches but use FHIR string search semantics (https://build.fhir.org/search.html#token)

In addition, there's this:
image

I could be wrong about this, but wanted to bring it up for discussion.

elsaperelli (changes requested):
Everything looks good to me except one thing that I think I got wrong when we implemented search in this [PR](https://github.com/projecttacoma/measure-repository-service/pull/3/commits/52c3fca5c3c49d90b288b1db367b030e4a76c660).

After some discussion, we think that `version` should be of type `token` and that it should not be treated as a string parameter like it is when it's grouped with `title`, `name`, and `description`. This is out of scope for this task, so maybe we should make a task to fix that! (If that is correct)

elsaperelli (approved):
After discussion at BL groom, whatever you decide for the version sounds good for now! Lgtm!

natjoe4 (approved):
Great work implementing and understanding this complex capability statement. Appreciate all the answers to my questions. Nice job :)

PR #15, opened 2023-02-01
Error for path parameter w/ query or fhir param

lmd59:
# Summary
Adds stricter checking for overlapping id fields. For both `$package` and `$data-requirements`, id can be given as part of the path itself, as in this path parameter: `4_0_1/Measure/:id/$data-requirements`, or the id may be passed as a query parameter (GET) or part of a FHIR parameters object used in a POST. If the id is in both the path and one of these other parameters, the server should throw a 400 bad request error.

## New behavior
Throws 400 bad request for redundant id input, even if the value of the id parameter is the same.

## Code changes

- Updates parameter checking in `package` function for `LibraryService` and `MeasureService`. This function is used for parameter checking in both the `$package` and `$data-requirements` operations. Uses `req.params.id` to determine whether the path parameter is used (Note: `args` cannot be used for this because `args` also collects query parameters).
- Adds tests for (Library, Measure) x ($package, $data-requirements) x (POST, GET), 6 total.

# Testing guidance
`npm run check`

For database setup:
- `npm run db:reset`

General testing:
- `npm start`
- Use insomnia or postman to test with various combinations of Library, Measure, data-requirements, package, POST, GET, and using path parameter or not.
- Example invalid GET: `http://localhost:3000/4_0_1/Measure/ColorectalCancerScreeningsFHIR/$package?id=ColorectalCancerScreeningsFHIR`
- Example invalid POST: `http://localhost:3000/4_0_1/Measure/ColorectalCancerScreeningsFHIR/$data-requirements` with fhir parameter object:
`{
  "resourceType": "Parameters",
  "parameter": [
    {
        "name": "id",
        "valueString": "ColorectalCancerScreeningsFHIR"
    },
    {
        "name": "periodStart",
        "valueDate": "2021-01-01"
    },
    {
        "name": "periodEnd",
        "valueDate": "2021-12-31"
    }
  ]
}`

### DIFF
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -5,7 +5,7 @@ import { Service } from '../types/service';
 import { createLibraryPackageBundle, createSearchsetBundle } from '../util/bundleUtils';
 import { BadRequestError, ResourceNotFoundError } from '../util/errorUtils';
 import { getMongoQueryFromRequest } from '../util/queryUtils';
-import { gatherParams, validateSearchParams } from '../util/validationUtils';
+import { gatherParams, validateSearchParams, validateParamIdSource } from '../util/validationUtils';
 
 const logger = loggers.get('default');
 
@@ -51,11 +51,7 @@ export class LibraryService implements Service<fhir4.Library> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
-    if (req.params.id && params.id) {
-      throw new BadRequestError(
-        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
-      );
-    }
+    validateParamIdSource(req.params.id, params.id);
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
--- a/src/services/LibraryService.ts
+++ b/src/services/LibraryService.ts
@@ -51,6 +51,11 @@ export class LibraryService implements Service<fhir4.Library> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
+    if (req.params.id && params.id) {
+      throw new BadRequestError(
+        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
+      );
+    }
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
--- a/src/services/MeasureService.ts
+++ b/src/services/MeasureService.ts
@@ -52,6 +52,11 @@ export class MeasureService implements Service<fhir4.Measure> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
+    if (req.params.id && params.id) {
+      throw new BadRequestError(
+        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
+      );
+    }
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
--- a/src/services/MeasureService.ts
+++ b/src/services/MeasureService.ts
@@ -5,7 +5,7 @@ import { Service } from '../types/service';
 import { createMeasurePackageBundle, createSearchsetBundle } from '../util/bundleUtils';
 import { BadRequestError, ResourceNotFoundError } from '../util/errorUtils';
 import { getMongoQueryFromRequest } from '../util/queryUtils';
-import { gatherParams, validateSearchParams } from '../util/validationUtils';
+import { gatherParams, validateSearchParams, validateParamIdSource } from '../util/validationUtils';
 import { Calculator } from 'fqm-execution';
 
 const logger = loggers.get('default');
@@ -52,11 +52,7 @@ export class MeasureService implements Service<fhir4.Measure> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
-    if (req.params.id && params.id) {
-      throw new BadRequestError(
-        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
-      );
-    }
+    validateParamIdSource(req.params.id, params.id);
     const id = args.id || params.id;
     const url = params.url;
     const version = params.version;
--- a/test/services/LibraryService.test.ts
+++ b/test/services/LibraryService.test.ts
@@ -317,6 +317,34 @@ describe('LibraryService', () => {
         });
     });
 
+    it('throws a 400 error when an id is included in both the path and a fhir parameter', async () => {
+      await supertest(server.app)
+        .post('/4_0_1/Library/testLibraryWithDeps/$package')
+        .send({ resourceType: 'Parameters', parameter: [{ name: 'id', valueString: 'testLibraryWithDeps' }] })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
+          );
+        });
+    });
+
+    it('throws a 400 error when an id is included in both the path and a query parameter', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library/testLibraryWithDeps/$package')
+        .query({ id: 'testLibraryWithDeps' })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
+          );
+        });
+    });
+
     it('throws a 404 error when both the library id and url are specified but one of them is invalid', async () => {
       await supertest(server.app)
         .post('/4_0_1/Library/$package')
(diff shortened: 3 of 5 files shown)


src/services/LibraryService.ts
```diff
@@ -51,6 +51,11 @@ export class LibraryService implements Service<fhir4.Library> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
+    if (req.params.id && params.id) {
+      throw new BadRequestError(
+        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
+      );
+    }
```
elsaperelli, line 58: Not a big deal for this task, but maybe we could put this into a function in `validationUtils.ts` since it's used twice? Or we leave it and keep this mind in the future if we need to use it again.

src/services/LibraryService.ts
```diff
@@ -51,6 +51,11 @@ export class LibraryService implements Service<fhir4.Library> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
+    if (req.params.id && params.id) {
+      throw new BadRequestError(
+        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 56:
```suggestion
        'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

src/services/MeasureService.ts
```diff
@@ -52,6 +52,11 @@ export class MeasureService implements Service<fhir4.Measure> {
     logger.info(`${req.method} ${req.path}`);
 
     const params = gatherParams(req.query, args.resource);
+    if (req.params.id && params.id) {
+      throw new BadRequestError(
+        'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 57:
```suggestion
        'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/LibraryService.test.ts
```diff
@@ -317,6 +317,34 @@ describe('LibraryService', () => {
         });
     });
 
+    it('throws a 400 error when an id is included in both the path and a fhir parameter', async () => {
```
elsaperelli, line 320:
```suggestion
    it('throws a 400 error when an id is included in both the path and a FHIR parameter', async () => {
```

test/services/LibraryService.test.ts
```diff
@@ -317,6 +317,34 @@ describe('LibraryService', () => {
...
+      await supertest(server.app)
+        .post('/4_0_1/Library/testLibraryWithDeps/$package')
+        .send({ resourceType: 'Parameters', parameter: [{ name: 'id', valueString: 'testLibraryWithDeps' }] })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 329:
```suggestion
            'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/LibraryService.test.ts
```diff
@@ -317,6 +317,34 @@ describe('LibraryService', () => {
...
+    it('throws a 400 error when an id is included in both the path and a query parameter', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Library/testLibraryWithDeps/$package')
+        .query({ id: 'testLibraryWithDeps' })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 343:
```suggestion
            'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/MeasureService.test.ts
```diff
@@ -326,6 +326,34 @@ describe('MeasureService', () => {
         });
     });
 
+    it('throws a 400 error when an id is included in both the path and a fhir parameter', async () => {
```
elsaperelli, line 329:
```suggestion
    it('throws a 400 error when an id is included in both the path and a FHIR parameter', async () => {
```

test/services/MeasureService.test.ts
```diff
@@ -326,6 +326,34 @@ describe('MeasureService', () => {
...
+      await supertest(server.app)
+        .post('/4_0_1/Measure/testWithRootLib/$package')
+        .send({ resourceType: 'Parameters', parameter: [{ name: 'id', valueString: 'testWithRootLib' }] })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 338:
```suggestion
            'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/MeasureService.test.ts
```diff
@@ -326,6 +326,34 @@ describe('MeasureService', () => {
...
+    it('throws a 400 error when an id is included in both the path and a query parameter', async () => {
+      await supertest(server.app)
+        .get('/4_0_1/Measure/testWithRootLib/$package')
+        .query({ id: 'testWithRootLib' })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 352:
```suggestion
            'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/MeasureService.test.ts
```diff
@@ -445,6 +473,34 @@ describe('MeasureService', () => {
           );
         });
     });
+
+    it('throws a 400 error when an id is included in both the path and a fhir parameter', async () => {
```
elsaperelli, line 477:
```suggestion
    it('throws a 400 error when an id is included in both the path and a FHIR parameter for $data-requirements operation', async () => {
```
I would add something like that at the end!

test/services/MeasureService.test.ts
```diff
@@ -445,6 +473,34 @@ describe('MeasureService', () => {
...
+      await supertest(server.app)
+        .post('/4_0_1/Measure/testWithRootLib/$data-requirements')
+        .send({ resourceType: 'Parameters', parameter: [{ name: 'id', valueString: 'testWithRootLib' }] })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 486:
```suggestion
            'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/MeasureService.test.ts
```diff
@@ -445,6 +473,34 @@ describe('MeasureService', () => {
...
+          );
+        });
+    });
+
+    it('throws a 400 error when an id is included in both the path and a query parameter', async () => {
```
elsaperelli, line 491:
```suggestion
    it('throws a 400 error when an id is included in both the path and a query parameter for $data-requirements operation', async () => {
```

test/services/MeasureService.test.ts
```diff
@@ -445,6 +473,34 @@ describe('MeasureService', () => {
...
+      await supertest(server.app)
+        .get('/4_0_1/Measure/testWithRootLib/$data-requirements')
+        .query({ id: 'testWithRootLib' })
+        .set('content-type', 'application/fhir+json')
+        .expect(400)
+        .then(response => {
+          expect(response.body.issue[0].code).toEqual('invalid');
+          expect(response.body.issue[0].details.text).toEqual(
+            'Id argument may not be sourced from both a path parameter and a query or fhir parameter.'
```
elsaperelli, line 500:
```suggestion
            'Id argument may not be sourced from both a path parameter and a query or FHIR parameter.'
```

test/services/LibraryService.test.ts
elsaperelli: Maybe add something to the end about it being for the $package operation. Same for the other description below

elsaperelli (changes requested):
Just some nit-picky comments. Otherwise nice! Works well :)

I did notice that there are 6 unit tests total since Library/$data-requirements has not been implemented yet if you want to change that in the PR description so we don't confuse our future selves :)

test/services/LibraryService.test.ts
lmd59: This is in the package operation test set, so you should end up with something like:
`LibraryService › $package › throws a 400 error when an id is included in both the path and a FHIR parameter`
(similar with below)

sarahmcdougall (approved):
Good work!

elsaperelli (approved):
Nice! :)

PR #16, opened 2023-02-06
ZOD (request param verification)

natjoe4:
# Summary
Added the ZOD dependency and implemented request input schemas for each of the defined operations

## New behavior
All existing request param input validation should continue to work, but now invalid params will throw 400 errors, unsupported params will throw 501 errors, and malformed params (url not a valid url) will throw 400 errors

## Code changes
 - Created request schemas for `search`, `$package`, and `$data-requirements` operations. Set all expected schema values to string, except for url for the time being.
 - Created dynamic functions to validate input and return proper OperationOutcome responses
 - Updated LibraryService and MeasureService files to use new requestSchemas
 - renamed `validationUtils` to `inputUtils` since all validation is now handled by Zod
 - Updated and added testing
 - Updated `.gitignore` to ignore eqcm-content-r4-2021 repo

# Testing guidance
 - `npm run db:reset`
 - `npm start`
 - Try all the requests in the included insomnia bundle and make sure all outcomes make sense
 - Check the schemas against the spec [here](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html) and make sure all params are accounted for.
 - NOTE: while `periodStart` and `periodEnd` are not currently listed as valid params for `$data-requirements` operations, we believe this is a mistake, and so they are included in the schemas

[ZOD_testing.json.zip](https://github.com/projecttacoma/measure-repository-service/files/10676504/ZOD_testing.json.zip)


src/requestSchemas.ts
```diff
@@ -0,0 +1,157 @@
...
+  .partial()
+  .strict()
+  .superRefine(catchInvalidParams([catchMissingIdentifyingInfo], UNSUPPORTED_PACKAGE_ARGS));
+
+export const DataRequirementsArgs = IdentifyingParameters.extend({
+  expression: z.string(),
+  parameters: z.string(),
+  manifest: z.string(),
+  periodStart: z.string(),
+  periodEnd: z.string(),
```
mgramigna, line 124: can probably refine these further to ensure they adhere to the FHIR date regex? e.g.

`z.string().regex(/([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1]))?)/)` (taken from FHIR spec)

could probably pull the regex out into a variable and reuse it

src/requestSchemas.ts
```diff
@@ -0,0 +1,157 @@
...
+  'include-dependencies': z.string(),
+  'include-components': z.string()
+})
+  .partial()
+  .strict()
+  .superRefine(catchInvalidParams([catchMissingIdentifyingInfo], UNSUPPORTED_DATA_REQ_ARGS));
+
+export const CoreSearchArgs = z
```
mgramigna, line 132: Do we even want to include parsing of search parameters that the measure repository service doesn't say are SHALLs or SHOULDs?

src/requestSchemas.ts
```diff
@@ -0,0 +1,157 @@
...
+export const IdentifyingParameters = z
+  .object({
+    id: z.string(),
+    url: z.string().url(),
+    version: z.string(),
+    identifier: z.string()
+  })
+  .partial();
+
+export const PackageArgs = IdentifyingParameters.extend({
```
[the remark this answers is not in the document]
mgramigna: Related: I actually realized a bug in the code unrelated to your PR. I would expect providing `include-terminology` as `'false'` in a request to not do the valueset resolution, but since everything comes through as a string over http, whenever the parameter is defined, the code doesn't care if its value is true or false. I think this can be solved in a couple ways, but maybe we can use a combo of [transform](https://zod.dev/?id=transform) and the actual output of the `parse` result.

For this specific case, something like 

```ts
  'include-terminology': z.enum(['true', 'false']).transform(val => val === 'true')
```

We could also try to to something similar for params that are integers, ensuring they are parsed as such. This will help us implement them later.

I also think that using the result of the parse is a good idea especially if we do this transform, as it will give us full type safety

src/requestSchemas.ts
```diff
@@ -0,0 +1,157 @@
...
+  capability: z.string(),
+  offset: z.string(),
+  count: z.string(),
```
mgramigna, line 107: see above comments about potential integer transforms

mgramigna (changes requested):
Looks good! I have a few small enhancement suggestions amongst other things

src/requestSchemas.ts
natjoe4: Yeah good thinking. Zod allows a .string().dateTime() type, but I chose not to use it because it rejects just `2019-01-01` style dates. regex seems like a great solution

src/requestSchemas.ts
natjoe4: My thinking is that keeping them here will cause the server to throw a 501 instead of a 400 for inclusion of these parameters, which seems slightly more accurate to me. Happy to make the change here if you disagree.

mgramigna: I guess it's a question of if we want to make a distinction between "We don't support this right now" and "we will never support this"

I think given the hard limit on resource types that will be supported, it's fine to include the unsupported ones I guess. Maybe there's a world where something changes and we actually want to add support for another search param or something. So it's fine to leave it in for a 501

test/services/MeasureService.test.ts
```diff
@@ -158,7 +158,7 @@ describe('MeasureService', () => {
     });
   });
 
-  describe('$package', () => {
+  describe.only('$package', () => {
```
mgramigna, line 161: treachery

```suggestion
  describe('$package', () => {
```

test/services/MeasureService.test.ts
```diff
@@ -292,22 +292,22 @@ describe('MeasureService', () => {
         });
     });
 
-    it('throws a 404 error when both the Measure id and url are specified but one of them is invalid', async () => {
+    it.only('throws a 404 error when both the Measure id and url are specified but one of them is invalid', async () => {
```
mgramigna, line 295: treachery

```suggestion
    it('throws a 404 error when both the Measure id and url are specified but one of them is invalid', async () => {
```

mgramigna (changes requested):

src/requestSchemas.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (!hasIdentifyingInfo(val)) {
+    ctx.addIssue({
+      code: z.ZodIssueCode.custom,
+      params: { serverErrorCode: constants.ISSUE.CODE.REQUIRED },
+      message: 'Must provide identifying information via either id, url, or identifier parameters'
+    });
+  }
+}
+
+const stringToBool = z.enum(['true', 'false']).transform(x => x === 'true');
```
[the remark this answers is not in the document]
natjoe4: It actually doesn't because `z.coerce.boolean()` just calls `Boolean(val)`, so if you sent in `false` it would actually evaluate to `true` since strings are truthy.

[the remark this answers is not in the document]
mgramigna: thanks I hate it

src/util/bundleUtils.ts
```diff
@@ -27,39 +28,71 @@ export function createSearchsetBundle<T extends fhir4.FhirResource>(entries: T[]
  * Takes in a measure resource, finds all dependent library resources and bundles them
  * together with the measure in a collection bundle
```
sarahmcdougall, line 29: We should update this since the function no longer takes in the measure

src/util/bundleUtils.ts
```diff
@@ -27,39 +28,71 @@ export function createSearchsetBundle<T extends fhir4.FhirResource>(entries: T[]
...
 
     return result;
   } else {
-    throw new BadRequestError(`Uploaded measure: ${measure.id} does not reference a Library`);
+    throw new BadRequestError(`Uploaded measure: ${measureForPackaging.id} does not reference a Library`);
   }
 }
 
 /**
  * Takes in a library resource, finds all dependent library resources and bundles them
```
sarahmcdougall, line 73: Should also update this doc string since we no longer use the library resource as input

test/util/inputUtils.test.ts
```diff
@@ -0,0 +1,29 @@
+import { gatherParams } from '../../src/util/inputUtils';
+
+const VALID_QUERY = { url: 'http://example.com' };
+
+const INVALID_QUERY = { invalid: 'test' };
```
sarahmcdougall, line 5:
```suggestion
```
This const was only used in `validateSearchParams` so we can get rid of it.

sarahmcdougall (changes requested):
Great work with this! And thank you for the detailed testing file!

Just a few small change suggestions but otherwise looks good!

mgramigna (changes requested):

src/requestSchemas.ts
```diff
@@ -0,0 +1,169 @@
...
+  })
+  .partial();
+
+export const PackageArgs = IdentifyingParameters.extend({
+  capability: z.string(),
+  offset: stringToNumber,
+  count: stringToNumber,
+  'system-version': z.string().url(),
+  'check-system-version': stringToBool,
+  'force-system-version': stringToBool,
```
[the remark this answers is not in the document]
mgramigna: Yes these should not be booleans, but `url` is also technically incorrect. These are type `canonical` in FHIR which supports things like `http://loinc.org|2.56`, which the `|` is not recognized as a valid URL.

if you want to create/use a regex for canonical, go for it, but otherwise anythiing that's canonical should just be `z.string()`

src/util/bundleUtils.ts
```diff
@@ -27,39 +28,71 @@ export function createSearchsetBundle<T extends fhir4.FhirResource>(entries: T[]
...
+  if (measure.length > 1) {
+    throw new BadRequestError(
+      `Multiple resources found in collection: Measure, with ${Object.keys(query)
+        .map(key => `${key}: ${query[key]}`)
+        .join(' and ')}. /Measure/$package operation must specify a single Measure`
+    );
+  }
+
+  const measureForPackaging = measure[0];
+  const includeTerminology = query['include-terminology'] ?? false;
```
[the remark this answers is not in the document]
mgramigna: Potentially solved by below comment about types + adding back the includeTerminology function param to the package function

src/requestSchemas.ts
```diff
@@ -0,0 +1,201 @@
...
+  if (!hasIdentifyingInfo(val)) {
+    ctx.addIssue({
+      code: z.ZodIssueCode.custom,
+      params: { serverErrorCode: constants.ISSUE.CODE.REQUIRED },
+      message: 'Must provide identifying information via either id, url, or identifier parameters'
+    });
+  }
+}
+
+const stringToBool = z.enum(['true', 'false']).transform(x => x === 'true');
```
mgramigna, line 105: I think I led you astray here. I tested all the edge cases, and look at this one:

```
{
  "resourceType": "Parameters",
  "parameter": [{
    "name": "id",
    "valueString": "ColorectalCancerScreeningsFHIR"
  }, {
    "name": "include-terminology",
    "valueBoolean": true
  }]
}
```

Gives the following error: `Expected 'true' | 'false', received boolean`

Issue is that `true/false` comes across as a string when encoded in the URL (reasonable), but as a boolean when in the params resource (also reasonable). We'll have to figure out some way to make both of them happy

mgramigna (changes requested):
sorry

src/requestSchemas.ts
mgramigna: Could probably use a `z.union` of the current schema and `z.boolean()`?

natjoe4: oof... good find. Could we just use a `z.union` here?

src/requestSchemas.ts
```diff
@@ -0,0 +1,201 @@
...
+  'predecessor',
+  'successor',
+  'topic'
+];
+
+const UNSUPPORTED_LIBRARY_SEARCH_ARGS = [...UNSUPPORTED_CORE_SEARCH_ARGS, 'content-type', 'type'];
+
+const UNSUPPORTED_MEASURE_SEARCH_ARGS = [...UNSUPPORTED_CORE_SEARCH_ARGS, 'publisher'];
+
+const customErrorMap: z.ZodErrorMap = (issue, ctx) => {
```
mgramigna, line 50: Discussed on a call, but this error map breaks things. @natjoe4 is looking into solutions

src/services/LibraryService.ts
```diff
@@ -22,9 +21,9 @@ export class LibraryService implements Service<fhir4.Library> {
     logger.info(`GET /Library`);
     const { query } = req;
     logger.debug(`Request Query: ${JSON.stringify(query, null, 2)}`);
-    validateSearchParams(query);
-    const parsedQuery = getMongoQueryFromRequest(query);
-    const entries = await findResourcesWithQuery<fhir4.Library>(parsedQuery, 'Library');
+    const parsedQuery = parseRequestSchema<typeof LibrarySearchArgs>(query, LibrarySearchArgs);
```
mgramigna, line 24: Isn't this generic inferred by TypeScript given that the generic is the type of the second parameter? You should be able to remove this

```suggestion
    const parsedQuery = parseRequestSchema(query, LibrarySearchArgs);
```

and it should still work. That's the beauty of generics!

src/services/LibraryService.ts
```diff
@@ -52,41 +51,11 @@ export class LibraryService implements Service<fhir4.Library> {
...
-      );
-    }
-    if (library.length > 1) {
-      throw new BadRequestError(
-        `Multiple resources found in collection: Library, with ${Object.keys(query)
-          .map(key => `${key}: ${query[key]}`)
-          .join(' and ')}. /Library/$package operation must specify a single Library`
-      );
-    }
+    const parsedParams = parseRequestSchema<typeof PackageArgs>({ ...params, ...query }, PackageArgs);
```
mgramigna, line 57: Same

```suggestion
    const parsedParams = parseRequestSchema({ ...params, ...query }, PackageArgs);
```

src/services/MeasureService.ts
```diff
@@ -23,9 +23,9 @@ export class MeasureService implements Service<fhir4.Measure> {
     logger.info(`GET /Measure`);
     const { query } = req;
     logger.debug(`Request Query: ${JSON.stringify(query, null, 2)}`);
-    validateSearchParams(query);
-    const parsedQuery = getMongoQueryFromRequest(query);
-    const entries = await findResourcesWithQuery<fhir4.Measure>(parsedQuery, 'Measure');
+    const parsedQuery = parseRequestSchema<typeof MeasureSearchArgs>(query, MeasureSearchArgs);
```
mgramigna, line 26: same

```suggestion
    const parsedQuery = parseRequestSchema(query, MeasureSearchArgs);
```

src/services/MeasureService.ts
```diff
@@ -53,42 +53,12 @@ export class MeasureService implements Service<fhir4.Measure> {
...
-      );
-    }
+    const query = extractIdentificationForQuery(args, params);
 
-    const query: Filter<any> = {};
-    if (id) query.id = id;
-    if (url) query.url = url;
-    if (version) query.version = version;
-    if (identifier) query.identifier = identifier;
+    const parsedParams = parseRequestSchema<typeof PackageArgs>({ ...params, ...query }, PackageArgs);
```
mgramigna, line 59: same

```suggestion
    const parsedParams = parseRequestSchema({ ...params, ...query }, PackageArgs);
```

src/services/MeasureService.ts
```diff
@@ -98,20 +68,26 @@ export class MeasureService implements Service<fhir4.Measure> {
    * requires parameters id and/or url and/or identifier, but also supports version as supplemental (optional)
    */
   async dataRequirements(args: RequestArgs, { req }: RequestCtx) {
+    const params = gatherParams(req.query, args.resource);
+    validateParamIdSource(req.params.id, params.id);
+    const query = extractIdentificationForQuery(args, params);
+
+    const parsedParams = parseRequestSchema<typeof MeasureDataRequirementsArgs>(
```
mgramigna, line 75: same

```suggestion
    const parsedParams = parseRequestSchema(
```

src/requestSchemas.ts
```diff
@@ -0,0 +1,210 @@
...
+      if (e.issues[0].code === z.ZodIssueCode.custom) {
+        if (e.issues[0].params?.type === 'NotImplemented') {
+          throw new NotImplementedError(e.issues[0].message);
+        } else {
+          throw new BadRequestError(e.issues[0].message, e.issues[0].params?.serverErrorCode);
+        }
+      } else if (e.issues[0].code === z.ZodIssueCode.unrecognized_keys) {
+        throw new BadRequestError(e.issues[0].message, constants.ISSUE.CODE.VALUE);
+      }
+      throw new BadRequestError(e.issues[0].message);
```
mgramigna, line 207: It looks like the default message for the incorrect type of a key is just "invalid input". Is there any way to make that better? I forget what they were like before. Maybe this is fine, but at the very least calling out the key that's in violation if that's available in the issue is a good call.

I sent `include-terminology=asdf` to violate the boolean schema for testing

mgramigna (changes requested):
We're so close I promise

src/requestSchemas.ts
mgramigna: This might also be specific to the union case. For example when I pass in a bad url the output is "invalid url" which is correct info

src/requestSchemas.ts
```diff
@@ -0,0 +1,210 @@
...
+export const LibrarySearchArgs = CoreSearchArgs.extend({
+  'content-type': z.string(),
+  type: z.string()
+})
+  .partial()
+  .strict()
+  .superRefine(catchInvalidParams([], UNSUPPORTED_LIBRARY_SEARCH_ARGS));
+
+export const MeasureSearchArgs = CoreSearchArgs.extend({
+  publisher: z.string()
```
sarahmcdougall, line 184: I see `publisher` listed for Library as well here: https://www.hl7.org/fhir/library.html#search
Should we move `publisher` to just appear under `CoreSearchArgs` instead?

src/requestSchemas.ts
```diff
@@ -0,0 +1,201 @@
...
+  'manifest',
+  'offset',
+  'system-version'
+];
+
+// Operation Definition: https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#requirements
+const UNSUPPORTED_DATA_REQ_ARGS = [
+  'expression',
+  'include-dependencies',
+  'include-components',
```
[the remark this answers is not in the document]
mgramigna: `:sort` in vim btw

src/requestSchemas.ts
```diff
@@ -0,0 +1,209 @@
...
+  'jurisdiction',
+  'predecessor',
+  'publisher',
+  'successor',
+  'topic'
+];
+
+const UNSUPPORTED_LIBRARY_SEARCH_ARGS = [...UNSUPPORTED_CORE_SEARCH_ARGS, 'content-type', 'type'];
+
+const UNSUPPORTED_MEASURE_SEARCH_ARGS = [...UNSUPPORTED_CORE_SEARCH_ARGS];
```
mgramigna, line 49: Need we this variable at all anymore?

mgramigna (approved):

sarahmcdougall (approved):
Thanks for bearing with all the change requests, just gave everything one last look-over and test, and everything looks good to merge!

PR #17, opened 2023-02-08
Expose $submit operation

sarahmcdougall:
# Summary
Exposes the `$submit` operation, which POSTs a new artifact in draft status.

## New Behavior
See [this portion of the spec](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#submit) for a brief description of the operation. Since the description is not detailed (and no OperationDefinition exists for this operation), we make the following assumptions:

* This operation is supported for both Measures and Libraries
* The operation is defined as a `POST` to `/(Library|Measure)/$submit` or `/(Library|Measure)/[id]/$submit`. Although the spec says that this can be a `PUT`, operations that involve state changes must be invoked by a POST. Therefore, we may not be able to use `PUT` for such an operation (and it is not exposed as an allowable option in node-fhir-server-core when setting up profile routes). Therefore, this PR deals with the `POST` scenario only, and we can re-visit the `PUT` case after getting more clarification on the spec.
* It is assumed that there are no IN params.
* It is assumed that the response should be `201` status since this is essentially a FHIR create with an additional check for draft status.
* For now, we assume the presence of dependent artifacts.

That being said, we expect that when a user `POST`s a Measure/Library resource to one of the specified endpoints with a FHIR Measure/Library in the body that is in “draft” status, the resource will be created and stored. When a user `POST`s a Measure/Library resource to one of the specified endpoints with a FHIR Measure/Library in the body that is not in “draft” status, a `BadRequest` error is thrown.

## Code Changes
* Documentation updates - README and capability statement
    * In capability statement: the operation `definition` must be defined, which is supposed to be an `OperationDefinition` canonical. There are currently no operation definitions for the authoring measure repository operations, so I put this as “[example.com](http://example.com/)” for now.
* New routes for both `MeasureService` and `LibraryService` in `serverConfig.ts`
* New `submit` function in both `MeasureService` and `LibraryService` - checks for correct content type header and resource type, checks for draft status, and then creates resource via `insertOne` on the collection. Returns 201 status and location of the resource
* Unit tests for submit operation and related helper functions

# Testing Guidance
* Run unit tests
* Send `POST` to `http://localhost:3000/4_0_1/Library/$submit`
    * With Library in “draft” status —> get 201
    * With Library not in “draft” status —> get 400
* Send `POST` to `http://localhost:3000/4_0_1/Measure/$submit`
    * With Measure in “draft” status —> get 201
    * With Measure not in “draft” status —> get 400
* Repeat with `http://localhost:3000/4_0_1/Library/<id>/$submit` and `http://localhost:3000/4_0_1/Measure/<id>/$submit`

### DIFF
--- a/scripts/dbSetup.ts
+++ b/scripts/dbSetup.ts
@@ -2,6 +2,7 @@ import { Connection } from '../src/db/Connection';
 import * as fs from 'fs';
 import * as dotenv from 'dotenv';
 import { MongoError } from 'mongodb';
+import { createResource } from '../src/db/dbOperations';
 
 dotenv.config();
 
@@ -92,16 +93,6 @@ async function insertFHIRModelInfoLibrary() {
   const fhirModelInfoLibrary: fhir4.Library = JSON.parse(fhirModelInfo); 
 
   await createResource(fhirModelInfoLibrary, 'Library');
-}
-
-/*
- * Inserts one data object into database with specified FHIR resource type
- */
-export async function createResource(data: fhir4.FhirResource, resourceType: string) {
-  const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
-  console.log(`Inserting ${resourceType}/${data.id} into database`);
-  await collection.insertOne(data);
-  return { id: data.id };
 }
 
 if (process.argv[2] === 'delete') {
--- a/scripts/dbSetup.ts
+++ b/scripts/dbSetup.ts
@@ -97,7 +97,7 @@ async function insertFHIRModelInfoLibrary() {
 /*
  * Inserts one data object into database with specified FHIR resource type
  */
-async function createResource(data: fhir4.FhirResource, resourceType: string) {
+export async function createResource(data: fhir4.FhirResource, resourceType: string) {
   const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
   console.log(`Inserting ${resourceType}/${data.id} into database`);
   await collection.insertOne(data);
--- a/src/config/capabilityStatementResources.json
+++ b/src/config/capabilityStatementResources.json
@@ -127,7 +127,7 @@
             "valueCode" : "SHALL"
           }],
           "name" : "submit",
-          "definition" : "example.com"
+          "definition" : "http://hl7.org/fhir/us/cqfmeasures/STU3/measure-repository-service.html#submit"
         }
       ]
     },
--- a/src/config/capabilityStatementResources.json
+++ b/src/config/capabilityStatementResources.json
@@ -267,7 +267,7 @@
             "valueCode" : "SHALL"
           }],
           "name" : "submit",
-          "definition" : "example.com"
+          "definition" : "http://hl7.org/fhir/us/cqfmeasures/STU3/measure-repository-service.html#submit"
         }
       ]
     }
--- a/src/config/capabilityStatementResources.json
+++ b/src/config/capabilityStatementResources.json
@@ -120,6 +120,14 @@
           }],
           "name" : "package",
           "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-package"
+        },
+        {
+          "extension" : [{
+            "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+            "valueCode" : "SHALL"
+          }],
+          "name" : "submit",
+          "definition" : "example.com"
         }
       ]
     },
@@ -252,6 +260,14 @@
           }],
           "name" : "data-requirements",
           "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Measure-data-requirements"
+        },
+        {
+          "extension" : [{
+            "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+            "valueCode" : "SHALL"
+          }],
+          "name" : "submit",
+          "definition" : "example.com"
         }
       ]
     }
--- a/src/db/dbOperations.ts
+++ b/src/db/dbOperations.ts
@@ -20,3 +20,13 @@ export async function findResourcesWithQuery<T extends fhir4.FhirResource>(
   const collection = Connection.db.collection(resourceType);
   return collection.find<T>(query, { projection: { _id: 0 } }).toArray();
 }
+
+/*
+ * Inserts one data object into database with specified FHIR resource type
+ */
+export async function createResource(data: fhir4.FhirResource, resourceType: string) {
+  const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
+  console.log(`Inserting ${resourceType}/${data.id} into database`);
+  await collection.insertOne(data);
+  return { id: data.id };
+}
--- a/src/db/dbOperations.ts
+++ b/src/db/dbOperations.ts
@@ -1,6 +1,8 @@
-import { FhirResourceType } from '@projecttacoma/node-fhir-server-core';
+import { loggers, FhirResourceType } from '@projecttacoma/node-fhir-server-core';
 import { Filter } from 'mongodb';
 import { Connection } from './Connection';
+
+const logger = loggers.get('default');
 
 /**
  * searches the database for the desired resource and returns the data
@@ -26,7 +28,7 @@ export async function findResourcesWithQuery<T extends fhir4.FhirResource>(
  */
 export async function createResource(data: fhir4.FhirResource, resourceType: string) {
   const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
-  console.log(`Inserting ${resourceType}/${data.id} into database`);
+  logger.info(`Inserting ${resourceType}/${data.id} into database`);
   await collection.insertOne(data);
   return { id: data.id };
 }
--- a/src/db/dbOperations.ts
+++ b/src/db/dbOperations.ts
@@ -1,5 +1,5 @@
 import { FhirResourceType } from '@projecttacoma/node-fhir-server-core';
-import { Filter, ObjectId } from 'mongodb';
+import { Filter } from 'mongodb';
 import { Connection } from './Connection';
 
 /**
@@ -19,10 +19,4 @@ export async function findResourcesWithQuery<T extends fhir4.FhirResource>(
 ) {
   const collection = Connection.db.collection(resourceType);
   return collection.find<T>(query, { projection: { _id: 0 } }).toArray();
-}
-
-export async function createResource(data: Partial<fhir4.FhirResource> & { _id: ObjectId }, resourceType: string) {
-  const collection = Connection.db.collection(resourceType);
-  await collection.insertOne(data);
-  return { id: data.id };
 }
--- a/src/db/dbOperations.ts
+++ b/src/db/dbOperations.ts
@@ -1,5 +1,5 @@
 import { FhirResourceType } from '@projecttacoma/node-fhir-server-core';
-import { Filter } from 'mongodb';
+import { Filter, ObjectId } from 'mongodb';
 import { Connection } from './Connection';
 
 /**
@@ -21,7 +21,7 @@ export async function findResourcesWithQuery<T extends fhir4.FhirResource>(
   return collection.find<T>(query, { projection: { _id: 0 } }).toArray();
 }
 
-export async function createResource(data: any, resourceType: string) {
+export async function createResource(data: Partial<fhir4.FhirResource> & { _id: ObjectId }, resourceType: string) {
   const collection = Connection.db.collection(resourceType);
   await collection.insertOne(data);
   return { id: data.id };
--- a/src/db/dbOperations.ts
+++ b/src/db/dbOperations.ts
@@ -26,18 +26,3 @@ export async function createResource(data: any, resourceType: string) {
   await collection.insertOne(data);
   return { id: data.id };
 }
-
-export async function updateResource(id: string, data: any, resourceType: string) {
-  const collection = Connection.db.collection(resourceType);
-  const results = await collection.replaceOne({ id }, data, { upsert: true });
-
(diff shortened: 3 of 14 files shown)


src/services/LibraryService.ts
```diff
@@ -89,4 +97,36 @@ export class LibraryService implements Service<fhir4.Library> {
...
+   * result of sending a POST request to:
+   * {BASE_URL}/4_0_1/Library/$submit or {BASE_URL}/4_0_1/Library/:id/$submit
+   * POSTs a new artifact in "draft" status. The operation results in an error if the artifact
+   * does not have status set to "draft."
+   */
+  async submit(args: RequestArgs, { req }: RequestCtx) {
+    logger.info(`${req.method} ${req.path}`);
+
+    const contentType: string | undefined = req.headers['content-type'];
+    checkContentTypeHeader(contentType);
```
elsaperelli, line 111: Out of scope for this task but just thought I'd note that it could be worth adding this function call to the other operations that send POST requests!

src/services/LibraryService.ts
```diff
@@ -89,4 +97,36 @@ export class LibraryService implements Service<fhir4.Library> {
...
+
+    const resource = req.body;
+    checkExpectedResourceType(resource.resourceType, 'Library');
+
+    // check for "draft" status on the resource
+    if (resource.status !== 'draft') {
+      throw new BadRequestError(`The artifact must be in 'draft' status.`);
+    }
```
elsaperelli, line 119: Probably very unnecessary, but I think this could be added to `validationUtils.ts` if you wanted!

src/services/LibraryService.ts
```diff
@@ -89,4 +97,36 @@ export class LibraryService implements Service<fhir4.Library> {
...
+
+    // lastUpdated should be second because it should overwrite a meta.lastUpdated tag in the request body
```
elsaperelli, line 121: I am a little confused about this comment...looks like it was leftover from the implementation of create in the test-server. Maybe clarify or just take it out?

src/services/MeasureService.ts
```diff
@@ -117,4 +125,36 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+
+    const resource = req.body;
+    checkExpectedResourceType(resource.resourceType, 'Measure');
+
+    // check for "draft" status on the resource
+    if (resource.status !== 'draft') {
+      throw new BadRequestError(`The artifact must be in 'draft' status.`);
+    }
+
+    // lastUpdated should be second because it should overwrite a meta.lastUpdated tag in the request body
```
elsaperelli, line 149: I am a little confused about this comment...looks like it was leftover from the implementation of create in the test-server. Maybe clarify or just take it out?

elsaperelli (changes requested):
Looks good to me besides some small comments!

Additionally, I know we discussed this but wanted to point it out on the PR for others. When I send a `GET` request to `http://localhost:3000/4_0_1/Library/$submit` and the body is the `ColorectalCancerScreeningsFHIR` Library, I get the following 500 repsonse:

image

I did some digging into why this was happening, but I couldn't find anything yet so I will continue to look into this. @sarahmcdougall noted that we do not have an issue with this Library at all in `deqm-test-server`.

src/config/capabilityStatementResources.json
```diff
@@ -120,6 +120,14 @@
...
           "name" : "package",
           "definition" : "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-package"
+        },
+        {
+          "extension" : [{
+            "url" : "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation",
+            "valueCode" : "SHALL"
+          }],
+          "name" : "submit",
+          "definition" : "example.com"
```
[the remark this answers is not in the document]
sarahmcdougall: Yeah I feel like there's a tradeoff here where if we include the description then we may forget to change it to the actual operation definition down the line (if it ever gets created), whereas example.com is more obviously incorrect. But at the same time I see what you're saying with this not giving any information to the capability statement. I'll include this definition and then I'll put a TODO comment in `serverConfig` about how the authoring measure repo definitions don't point to operation definitions so that we remember to change them in the future

src/services/LibraryService.ts
sarahmcdougall: I'll hold of on that for now since this is just a simple if statement, but if we end up checking draft status across several other operations then we can consider moving this into a helper

src/services/LibraryService.ts
sarahmcdougall: It's basically saying that when we update `resource.meta` we should take the existing contents of `resource.meta` and then overwrite the `lastUpdatedTag` that currently exists on `resource.meta`.

However, I just looked the [spec](http://www.hl7.org/fhir/resource-definitions.html#Resource.meta) and saw this: "This value [of `lastUpdated`] is always populated except when the resource is first being created." So I think this can be removed entirely (and will probably need to be removed from deqm test server at some point because it appears in the `baseCreate` fn when it should only be in `baseUpdate`)

sarahmcdougall:
> Additionally, I know we discussed this but wanted to point it out on the PR for others. When I send a `GET` request to `http://localhost:3000/4_0_1/Library/$submit` and the body is the `ColorectalCancerScreeningsFHIR` Library, I get the following 500 repsonse:

I updated the body parser limit to 50mb so if you try POSTing the resource again it should (hopefully) work this time

src/services/MeasureService.ts
```diff
@@ -117,4 +131,33 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+
+    // check for "draft" status on the resource
+    if (resource.status !== 'draft') {
+      throw new BadRequestError(`The artifact must be in 'draft' status.`);
+    }
+
+    const res = req.res;
+    // create new resource with server-defined id
+    resource['id'] = uuidv4();
+    await createResource(resource, 'Library');
```
natjoe4, line 158:
```suggestion
    await createResource(resource, 'Measure');
```

src/services/MeasureService.ts
```diff
@@ -117,4 +131,33 @@ export class MeasureService implements Service<fhir4.Measure> {
...
+    res.status(201);
+    const location = `${constants.VERSIONS['4_0_1']}/Library/${resource.id}`;
```
natjoe4, line 160:
```suggestion
    const location = `${constants.VERSIONS['4_0_1']}/Measure/${resource.id}`;
```

src/util/validationUtils.ts
```diff
@@ -48,3 +48,23 @@ export function validateParamIdSource(pathId: any, paramId: any) {
     );
   }
 }
+
+/**
+ * Checks that the content-type from the request headers accepts json + fhir.
+ */
+export function checkContentTypeHeader(contentType: string | undefined) {
```
natjoe4, line 55:
```suggestion
export function checkContentTypeHeader(contentType?: string) {
```
nitpicking here, but I think this is a bit more to our standard

test/services/LibraryService.test.ts
```diff
@@ -385,6 +385,48 @@ describe('LibraryService', () => {
         });
     });
   });
+
+  describe('$submit', () => {
+    it('returns 201 status with populated content location when provided correct headers and FHIR Library', async () => {
+      await supertest(server.app)
+        .post('/4_0_1/Library/$submit')
+        .send({resourceType: 'Library', status: 'draft'})
+        .set('Accept', 'application/json+fhir')
```
natjoe4, line 394: Are we ever planning to use this Accept header? If not, I'd remove it from the unit tests

src/services/LibraryService.ts
```diff
@@ -89,4 +102,33 @@ export class LibraryService implements Service<fhir4.Library> {
...
+      throw new BadRequestError(`The artifact must be in 'draft' status.`);
+    }
+
+    const res = req.res;
+    // create new resource with server-defined id
+    resource['id'] = uuidv4();
+    await createResource(resource, 'Library');
+    res.status(201);
+    const location = `${constants.VERSIONS['4_0_1']}/Library/${resource.id}`;
+    res.set('Content-Location', `${process.env.HOST}:${process.env.PORT}/${location}`);
```
natjoe4, line 132: for this `content-location` header, I'm guessing we want either the whole url `http://localhost:3000/4_0_1/Library/:id` or just the endpoint `/4_0_1/Library/:id`. I'd lean towards the latter since that's what we do in deqm-test-server. Also unclear if this should be `Content-Location` or just `Location` like it is in the test server for simple resource creates. Ugh... so many unknowns with this stuff.

natjoe4: It also looks like the headers get lowercased when they go back through express. Does it make sense to change this to just `content-location`?

natjoe4 (changes requested):
Great work with little info lol. Mainly small changes and a question: Do we want to do resource validation at all on the bodies uploaded for $submit operations? Right now we allow anything to be passed-in and stored, which I could see having issues with type-safety, when we reach into the Library collection and pull out something with invalid Library fields. Maybe just a new task though?

src/config/capabilityStatementResources.json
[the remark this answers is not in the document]
elsaperelli: 🤝

scripts/dbSetup.ts
```diff
@@ -97,7 +97,7 @@ async function insertFHIRModelInfoLibrary() {
 /*
  * Inserts one data object into database with specified FHIR resource type
  */
-async function createResource(data: fhir4.FhirResource, resourceType: string) {
+export async function createResource(data: fhir4.FhirResource, resourceType: string) {
```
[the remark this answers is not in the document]
elsaperelli: I agree now that it is being used in other places!

src/services/LibraryService.ts
sarahmcdougall: This snippet is what happens in the `node-fhir-server-core` `create`:
Screen Shot 2023-02-13 at 10 35 12 AM

Looks like both `Content-Location` and `Location` get set, so maybe we should set both?

natjoe4: Yeah I think both seems like the safest option :)

mgramigna: deqm-test-server only uses `Location`. The Content-Location appears to only be for version-specific reads which we don't support. Location in deqm-test-server appears to only be the slug, not the full url

mgramigna: Actually, idk where they even got `Content-Location` from. See FHIR docs, says to use `Location` for both: https://build.fhir.org/http.html#create

src/db/dbOperations.ts
```diff
@@ -20,3 +20,13 @@ export async function findResourcesWithQuery<T extends fhir4.FhirResource>(
   const collection = Connection.db.collection(resourceType);
   return collection.find<T>(query, { projection: { _id: 0 } }).toArray();
 }
+
+/*
+ * Inserts one data object into database with specified FHIR resource type
+ */
+export async function createResource(data: fhir4.FhirResource, resourceType: string) {
+  const collection = Connection.db.collection<fhir4.FhirResource>(resourceType);
+  console.log(`Inserting ${resourceType}/${data.id} into database`);
```
natjoe4, line 29: I think we probably want this as a `logger.info` instead

natjoe4 (changes requested):
Great work! One tiny comment

test/services/LibraryService.test.ts
```diff
@@ -385,6 +385,45 @@ describe('LibraryService', () => {
...
+        .post('/4_0_1/Library/$submit')
+        .send({resourceType: 'Library', status: 'draft'})
+        .set('content-type', 'application/json+fhir')
+        .expect(201)
+        .then(response => {
+          expect(response.headers.location).toBeDefined();
+        });
+    });
+
+    it('returns 201 status with populated content location when id is represent in the path', async () => {
```
elsaperelli, line 401:
```suggestion
    it('returns 201 status with populated location when id is represent in the path', async () => {
```

test/services/LibraryService.test.ts
```diff
@@ -385,6 +385,45 @@ describe('LibraryService', () => {
         });
     });
   });
+
+  describe('$submit', () => {
+    it('returns 201 status with populated content location when provided correct headers and FHIR Library', async () => {
```
elsaperelli, line 390:
```suggestion
    it('returns 201 status with populated location when provided correct headers and FHIR Library', async () => {
```

elsaperelli: Not sure if this is necessary actually, your call

test/services/MeasureService.test.ts
```diff
@@ -503,6 +503,44 @@ describe('MeasureService', () => {
     });
   });
 
+  describe('$submit', () => {
+    it('returns 201 status with populated content location when provided correct headers and FHIR Measure', async () => {
```
elsaperelli, line 507:
```suggestion
    it('returns 201 status with populated location when provided correct headers and FHIR Measure', async () => {
```

test/services/MeasureService.test.ts
```diff
@@ -503,6 +503,44 @@ describe('MeasureService', () => {
...
+        .post('/4_0_1/Measure/$submit')
+        .send({resourceType: 'Measure', status: 'draft'})
+        .set('content-type', 'application/json+fhir')
+        .expect(201)
+        .then(response => {
+          expect(response.headers.location).toBeDefined();
+        });
+    });
+
+    it('returns 201 status with populated content location when id is represent in the path', async () => {
```
elsaperelli, line 518:
```suggestion
    it('returns 201 status with populated location when id is represent in the path', async () => {
```

elsaperelli (changes requested):
Great job with this PR! You found a great solution to that 500 response very quickly. Just a couple very nit-picky things and then should be good to go!

natjoe4 (approved):
Looks good 👍

elsaperelli (approved):

PR #22, opened 2023-03-09
remove periodStart and periodEnd parameters from Library data-require…

lmd59:
…ments

# Summary
Changed parameter checking to only allow periodStart and periodEnd for the Measure endpoint, and not the Library endpoint.

## New behavior
Server throws a 400 error when periodStart or periodEnd are included in the parameters for the Library data-requirements endpoint.

## Code changes
- Update to README to specify that periodStart and periodEnd are only applicable for the measure endpoint.
- Updated LibraryService no longer uses period in fqm-execution data-requirements call
- requestSchemas.ts moves period checking from CommonDataRequirementsArgs to MeasureDataRequirementsArgs
- Tests updated to reflect allowed parameters and specifically check for 400 for periodStart and periodEnd

# Testing guidance
See testing guidance for: https://github.com/projecttacoma/measure-repository-service/pull/18
Any request that has a period in the parameters should throw a 400. If you then delete the periodStart and periodEnd from the parameters, the request should work as expected (200 and appropriate library body returned).

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ The Measure Repository Service server supports the `Measure` and `Library` `$dat
 - version
 - identifier
 
-Supported optional parameters are:
+Supported optional parameters for `Measure` are:
 
 - periodStart
 - periodEnd
--- a/src/requestSchemas.ts
+++ b/src/requestSchemas.ts
@@ -139,16 +139,18 @@ export const CommonDataRequirementsArgs = IdentifyingParameters.extend({
   'include-dependencies': stringToBool,
   manifest: z.string(),
   parameters: z.string(),
-  periodEnd: checkDate,
-  periodStart: checkDate,
   'system-version': z.string()
 })
   .partial()
   .strict();
 
-export const MeasureDataRequirementsArgs = CommonDataRequirementsArgs.superRefine(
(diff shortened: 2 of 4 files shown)


sarahmcdougall (approved):
Looks great and works great! No suggested changes from me :)

This didn't show up in the code diff, but I noticed on [line 453](https://github.com/projecttacoma/measure-repository-service/blob/6028f18d924404b719421c794d691769d2002c34/test/services/LibraryService.test.ts#L453) of LibraryService.test.ts we say "and no period params." Thoughts on removing this since we no longer need to explicitly state it (since we should never expect these params to be present in the first place)?

elsaperelli (changes requested):
I agree with Sarah's change suggestion, but other than that looks good to me!

lmd59:
> Looks great and works great! No suggested changes from me :)
> 
> This didn't show up in the code diff, but I noticed on [line 453](https://github.com/projecttacoma/measure-repository-service/blob/6028f18d924404b719421c794d691769d2002c34/test/services/LibraryService.test.ts#L453) of LibraryService.test.ts we say "and no period params." Thoughts on removing this since we no longer need to explicitly state it (since we should never expect these params to be present in the first place)?

Good call- updated!

elsaperelli (approved):

PR #26, opened 2023-04-07
Resource counts

lmd59:
# Summary

## Code changes

# Testing guidance


frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,86 @@
...
+      </div>
+    </div>
+  );
+}
+
+/**
+ * Serverside props pulls id data of a certain resourceType to pass to the page before it's sent to browser
+ * @returns props for the [resourceType] page that pass resourceType and ids of resources of that type
+ */
+export const getServerSideProps: GetServerSideProps = async context => {
```
Comment (elsaperelli) [inline frontend/pages/[resourceType].tsx:77]: Do we want to have a loading spinner of some kind while we are fetching the resource data?

frontend/components/ResourceCounts.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }, []);
+
+  /**
+   * Returns the resource counts key:value pairs object as an array of buttons
+   * @returns array of JSX Buttons that are resources and their counts
+   */
+  const ResourceButtonsGroup = () => {
+    const buttonArray = Object.keys(resources).map(resourceType => (
+      <Link href={`/${resourceType}`} key={resourceType}>
+        <Button
```
elsaperelli, line 37: Similar to my other comment, but I think if we made the button the width of the resources nav bar and also put some white space between the button on the left side it might look better? This is what it current looks like:

Screenshot 2023-04-07 at 3 14 59 PM

elsaperelli (changes requested):
Amazing job with this! Frontend dev woo! Just a couple comments/suggestions

frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,86 @@
...
+            border: '1px solid',
+            borderColor: '#DEE2E6',
+            borderRadius: '20px',
+            marginTop: '10px',
+            marginBottom: '20px',
+            marginLeft: '150px',
+            marginRight: '150px'
+          }}
+        >
+          <ul>{idItems}</ul>
```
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply frontend/pages/[resourceType].tsx:66]: I think this is a good idea! I was thinking we could either disable the button for a resource type if no resources are available, or add some text on the page like Elsa said. Something simple like "There are no measures on the server"

frontend/components/ResourceCounts.tsx
```diff
@@ -0,0 +1,67 @@
...
+    // Note: count may not update without refresh if there ends up being a way to change it on the server through the app
+    Promise.all([
+      fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/Measure`),
+      fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/Library`)
+    ])
+      .then(([resMeasure, resLibrary]) => Promise.all([resMeasure.json(), resLibrary.json()]))
+      .then(([measureBundle, libraryBundle]) => {
+        setResources({ Measure: measureBundle.total || 0, Library: libraryBundle.total || 0 });
+      })
+      .catch(error => {
```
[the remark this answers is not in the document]
sarahmcdougall: I also noticed this (I was using an incompatible node version and so I never connected to the mongo database and was wondering why there were no resources loaded 🤦‍♀️ lol) I think adding the Mantine notification is a good idea and probably won't be a lot of overhead to implement since we already have it implemented in the deqm-test-server-frontend

sarahmcdougall (commented):
Elsa beat me to my comments 😛 I only have one additional comment: I uploaded libraries from a few different measures and noticed that the styled `div` is not dynamically sized to the amount of ids we are displaying. I think in deqm-test-server-frontend we use pagination so this is not an issue. Just wanted to see what folks think... since this is a demo app, will we ever have this many resources of a single type on the server? Should we remove this `div`, make its size more dynamic, keep as-is until we update the display with the library cards?
mrs-frontend-library

frontend/pages/[resourceType].tsx
Comment (mgramigna) [reply frontend/pages/[resourceType].tsx:77]: Loading spinners only make sense during client side fetching. With `getServerSideProps`, nothing is rendered on the page until the resource data is returned. This function runs on the server before returning the HTML that gets sent back to the client, so there will be no loading state

Comment (mgramigna) [inline frontend/pages/[resourceType].tsx:77]: see [this comment](https://github.com/projecttacoma/measure-repository/pull/28#discussion_r1165607014) and [this one](https://github.com/projecttacoma/measure-repository/pull/28#discussion_r1165607014)

lmd59:
> * npm run db:loadBundle "ecqm-content-r4-2021/bundles/measure/ColorectalCancerScreeningsFHIR/ColorectalCancerScreeningsFHIR-bundle.json"

I ended up just getting rid of the 500px styling - do you think that works?

frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,86 @@
...
+            color="cyan"
+            radius="md"
+            size="md"
+            variant="subtle"
+            style={{
+              padding: '2px'
+            }}
+          >
+            <div>
+              {resourceType}/{id}
```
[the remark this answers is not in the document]
Comment (lmd59) [reply frontend/pages/[resourceType].tsx:24]: Made the button take up the whole width!

frontend/pages/[resourceType].tsx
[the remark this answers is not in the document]
Comment (lmd59) [reply frontend/pages/[resourceType].tsx:66]: Added text for when no resources are available!

frontend/components/ResourceCounts.tsx
lmd59: Made button take up whole width!

frontend/components/ResourceCounts.tsx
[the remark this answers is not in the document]
lmd59: I have an initial take on this, which shows the error in the resources section instead of showing the 0 count resources. I'm mixed on it as an approach and would be interested in opinions.

frontend/components/ResourceCounts.tsx
```diff
@@ -0,0 +1,67 @@
+import { useEffect, useState } from 'react';
+import { Badge, Button, Stack } from '@mantine/core';
+import Link from 'next/link';
+
+/**
+ * Component which retrieves all resources and their counts and translates them into buttons
+ * @returns array of JSX Buttons
+ */
+const ResourceCounts = () => {
+  // set initial state to no resources
```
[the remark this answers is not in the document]
lmd59: This is addressed by error solution above. I'm going to resolve this at the moment in favor of conversation on that [thread](https://github.com/projecttacoma/measure-repository/pull/26#discussion_r1167222131), but I can open it up again if we don't think that solution makes sense.

frontend/pages/[resourceType].tsx
Comment (lmd59) [reply frontend/pages/[resourceType].tsx]: Did some error throwing to reduce the complexity of types since we know a certain amount about the types that we'll be calling for and getting back from the server.

lmd59:
Note: `npm install` after recent rebase!

frontend/components/ResourceCounts.tsx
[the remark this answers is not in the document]
lmd59: Added a bit of styling. Basic look:
Screen Shot 2023-04-16 at 2 28 12 PM

frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,107 @@
...
+  );
+}
+
+/**
+ * Serverside props pulls id data of a certain resourceType to pass to the page before it's sent to browser
+ * @returns props for the [resourceType] page that pass resourceType and ids of resources of that type
+ */
+export const getServerSideProps: GetServerSideProps<{
+  ids: (string | undefined)[];
+  resourceType: string;
```
Comment (elsaperelli) [inline frontend/pages/[resourceType].tsx:88]: Would it be worth it to change resourceType from type string to something more specific like `'Measure' | 'Library'`?

frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,107 @@
...
+          })}
+        >
+          <div>
+            {resourceType}/{id}
+          </div>
+        </Button>
+      </Link>
+    );
+  });
+  return (
```
Comment (elsaperelli) [inline frontend/pages/[resourceType].tsx:35]: I would add a back button on this page so the user can return to the home page!

elsaperelli (changes requested):
Great changes! I think the addition of the server error is very helpful and I love the placement of it in the nav bar- good call! Just some small inline comments & questions!

frontend/pages/[resourceType].tsx
Comment (lmd59) [reply frontend/pages/[resourceType].tsx:88]: I def considered that. It's a bit of a tossup, but I figured in case we wanted to have MeasureReports in here some day I might leave it more flexible for now. *shrug* could go either way lol

frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,107 @@
...
+    <div
+      style={{
+        width: '78vw'
+      }}
+    >
+      <Grid columns={7}>
+        <Grid.Col offset={3} span={2} style={{ paddingTop: '5px' }}>
```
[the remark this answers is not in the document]
Comment (lmd59) [reply frontend/pages/[resourceType].tsx:42]: Huh- okay! Coolios :-)

frontend/components/ResourceCounts.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [resources, setResources] = useState<Record<string, number>>({
+    Measure: 0,
+    Library: 0
+  });
+  const [errorMessage, setErrorMessage] = useState<string>('');
```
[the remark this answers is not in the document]
lmd59: I think that if another error message were to be used in the future, you'd probably want to have it show up similarly. Since this is handling state for rendering components, I would say that any error message could be set in this state as long as you want it to show up out of this component.

frontend/components/ResourceCounts.tsx
```diff
@@ -0,0 +1,98 @@
...
+            color: '#E00000',
+            border: '1px solid',
+            borderColor: '#DEE2E6',
+            borderRadius: '20px',
+            marginLeft: '15px',
+            marginRight: '15px'
+          }}
+        >
+          {' '}
+          Resources could not be displayed due to an error retrieving them from the server.{' '}
```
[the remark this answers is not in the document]
lmd59: Also thought these were weird. I thought prettier had added them, but doesn't seem to be doing so now, so maybe they were leftover cruft from a different approach. Thanks!

frontend/components/ResourceCounts.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { Badge, Button, Stack } from '@mantine/core';
+import Link from 'next/link';
+
+/**
+ * Component which retrieves all resources and their counts and translates them into buttons
+ * @returns array of JSX Buttons
+ */
+const ResourceCounts = () => {
+  // set initial state to no resources
+  const [resources, setResources] = useState<Record<string, number>>({
```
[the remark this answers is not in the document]
lmd59: Honestly, the first load with 0 kind of bothered me anyway in addition to this edge case of the fast clicker. lol. Now setting state to -1 initially and only showing the Link if valid.

frontend/pages/[resourceType].tsx
Comment (lmd59) [reply frontend/pages/[resourceType].tsx:35]: Added back button and homepage link for top title.

frontend/pages/[resourceType].tsx
Comment (sarahmcdougall) [reply frontend/pages/[resourceType].tsx:92]: I created a [util file](https://github.com/projecttacoma/measure-repository/blob/main/frontend/util/types/fhir.ts) when implementing the resource JSON rendering, so we could add to that if we really wanted to and create a custom type for just Measure and Library strings. But I think it's fine to stay flexible for now. I don't anticipate us needing to constrain `resourceType` to certain values since it's assumed that the server will already restrict what resource types can exist on the server

Comment (mgramigna) [reply frontend/pages/[resourceType].tsx:92]: `export type ArtifactResourceType = FhirArtifact["resourceType"];` would be a nice dynamic solution

frontend/pages/[resourceType].tsx
```diff
@@ -0,0 +1,111 @@
...
+      <Grid columns={7}>
+        <Grid.Col offset={3} span={2} style={{ paddingTop: '6px' }}>
+          <h2 style={{ color: 'gray', marginTop: '0px', marginBottom: '8px' }}>{`${resourceType} IDs`}</h2>
+        </Grid.Col>
+        <Grid.Col
+          span={2}
+          style={{
+            paddingTop: '6px'
+          }}
+        ></Grid.Col>
```
Comment (sarahmcdougall) [inline frontend/pages/[resourceType].tsx:54]: Now that we have the back button, the dividers (for resource counts and the resourceType page) are no longer aligned. I think we might want to put the back button component in a `Grid.Col` so that it is aligned with the header without affecting the location of the divider.

Played around with this a little, and this achieved what I was looking for (not sure if it's the most efficient... I'm not the best with grids):

```javascript
<Grid columns={7}>
  <Grid.Col offset={0} span={1}>
    <div >
      <BackButton />
    </div>
  </Grid.Col>
  <Grid.Col offset={2} span={2} style={{ paddingTop: '6px' }}>
    <h2 style={{ color: 'gray', marginTop: '0px', marginBottom: '8px' }}>{`${resourceType} IDs`}</h2>
  </Grid.Col>
</Grid>
<Divider my="md" style={{ marginTop: '14px' }} />
```

sarahmcdougall (changes requested):
Great work implementing all the suggested changes - looks great! One additional comment from me:

frontend/pages/[resourceType].tsx
Comment (lmd59) [reply frontend/pages/[resourceType].tsx:54]: Did not notice that at all- good catch!

elsaperelli (approved):
Great job! Lgtm

sarahmcdougall (approved):
Great changes!

PR #28, opened 2023-04-11
Render JSON content for selected resource

sarahmcdougall:
# Summary
Adds dynamic route for rendering the JSON content of a selected resource on the server.

## New behavior
We now have a new page that has tabs for the JSON, ELM, CQL, and narrative content of a selected resource (all are disabled except for JSON for now - this will change once the other tab functionality is implemented). The JSON tab renders the content of the selected resource.

## Code changes
* New component `BackButton.tsx` to go back to the previous page (will bring the user back to the list of resources once that is merged)
* New page [id].tsx that renders the JSON content of the selected resource. Uses `getServerSideProps` to fetch the JSON content and stringify it. Uses a Mantine scrollable `Prism` to display the JSON content

# Testing guidance
* Create a local `.env.local` file in the frontend directory with the `NEXT_PUBLIC_MRS_SERVER` environment variable. You can use `.env.example` as an example
* Run `npm run start:all` to start the server and frontend (backend defaults to 3000, frontend defaults to 3001)
* Navigate to the url for the frontend app to access the application.
* **Note**: The resource list page from https://github.com/projecttacoma/measure-repository/pull/26 is not merged into main. To access the rendered JSON page, type the url into the browser box directly. The url should be of the form `<localhost url>/<resourceType>/<id>`. For example, if the `ColorectalCancerScreeningsFHIR` measure is on the server, navigate to `http://localhost:3001/Measure/ColorectalCancerScreeningsFHIR` and Nextjs will resolve the route.
* Ensure the JSON content properly loads.

For database loading (with an example measure bundle):
```
npm run db:reset
npm run db:loadBundle "ecqm-content-r4-2021/bundles/measure/ColorectalCancerScreeningsFHIR/ColorectalCancerScreeningsFHIR-bundle.json"
```


frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Serverside props pulls JSON data of a specified resource to pass to the page before it's sent to browser
+ * @returns props for the [resourceType]/[id] page that pass JSON data of a specified resource
+ */
+export const getServerSideProps: GetServerSideProps = async context => {
+  const { resourceType, id } = context.query;
+  // Fetch resource data
+  const res = await fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/${resourceType}/${id}`);
+  const resourceJson = await res.json();
+  const jsonString = JSON.stringify(resourceJson, null, 2);
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:72]: Should consider not stringifying the prop on the server and just passing up the whole FHIR resource. Other tabs will need to access content on the resource itself. You can probably type this as `fhir4.Measure | fhir4.Library` as well.

Prob need to install `@types/fhir` on the frontend in that case.

Comment (sarahmcdougall) [reply frontend/pages/[resourceType]/[id].tsx]: Good point, just made that change 👍

elsaperelli (approved):
Lgtm again!

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+ * Serverside props pulls JSON data of a specified resource to pass to the page before it's sent to browser
+ * @returns props for the [resourceType]/[id] page that pass JSON data of a specified resource
+ */
+export const getServerSideProps: GetServerSideProps = async context => {
+  const { resourceType, id } = context.query;
+  // Fetch resource data
+  const res = await fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/${resourceType}/${id}`);
+  const resourceJson = await res.json();
+  // pass JSON data to the page via props
+  return { props: { jsonData: resourceJson as fhir4.Measure | fhir4.Library } };
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:73]: Personal preference, but I would type the result of res.json directly as opposed to in the return:

```suggestion
  const resourceJson = (await res.json()) as fhir4.Library | fhir4.Measure;
  // pass JSON data to the page via props
  return { props: { jsonData: resourceJson } };
```

This way we have the type safety as early as possible, which will be useful in a world where we need to do something with the resource on the server before passing up the prop.

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+export default function ResourceIDPage({ jsonData }: { jsonData: fhir4.Measure | fhir4.Library }) {
+  const router = useRouter();
+  const { resourceType, id } = router.query;
+  return (
+    <div>
+      <Stack spacing="xs">
+        <div
+          style={{
+            float: 'left'
+          }}
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:21]: I don't think this does anything

```suggestion
```

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          <Group>
+            <BackButton />
+            <Text size="xl" weight={700} color="gray">
+              {resourceType}/{id}
+            </Text>
+          </Group>
+        </div>
+        <Divider my="sm" style={{ paddingBottom: '5px' }} />
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:30]: Another nitpick but it's recommended to have padding and other spacing be consistent and scalable, which usually means multiples of 2 or 4. I usually go off of [TailwindCSS' padding system](https://tailwindcss.com/docs/padding) even though we're not using that library here (sadly). Note how a lot of their classes go from 2px, 4px, 6px, 8px, 10, 12, etc.

I'd either make this 4 or 6 instead

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+            <Tabs.Tab value="narrative" disabled>
+              Narrative
+            </Tabs.Tab>
+          </Tabs.List>
+
+          <Tabs.Panel value="json" pt="xs">
+            <ScrollArea>
+              <Prism
+                language="json"
+                data-testid="prism-page-content"
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:49]: don't need this
```suggestion
```

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+                colorScheme="light"
+                style={{ maxWidth: '78vw', height: '80vh', backgroundColor: '#FFFFFF' }}
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:51]: Is there a reason we're restricting this to `78vw` and `80vh`? I don't see a reason why for now the prsim viewer can't just take up the whole screen, which you could accomplish by just deleting this whole style prop.

The content is scrollable and seems fine. This style as is adds a lot of whitespace to the right where there will never really be content at least right now

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+            <Tabs.Tab value="cql" disabled>
+              CQL
+            </Tabs.Tab>
+            <Tabs.Tab value="narrative" disabled>
+              Narrative
+            </Tabs.Tab>
+          </Tabs.List>
+
+          <Tabs.Panel value="json" pt="xs">
+            <ScrollArea>
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:46]: Looks like the `Prism` component already comes with a scroll area component, so I think we should delete this. This also fixes the copy button getting pushed way off screen when deleting the below style

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,74 @@
...
+import BackButton from '../../components/BackButton';
+
+/**
+ * Component which displays the JSON/ELM/CQL/narrative content of an individual resource using
+ * Mantine tabs
+ * @returns JSON/ELM/CQL/narrative content of the individual resource in a Prism component
+ */
+export default function ResourceIDPage({ jsonData }: { jsonData: fhir4.Measure | fhir4.Library }) {
+  const router = useRouter();
+  const { resourceType, id } = router.query;
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:14]: Now that we have the type safety of the resource, it's probably best to remove these entirely and just render `jsonData.resourceType` and `jsonData.id` instead. That way we don't have to wait for the client to catch up and parse the router query in order to render that information on the page.

mgramigna (changes requested):

frontend/components/BackButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <Button
+      data-testid="back-button"
+      onClick={() => router.back()}
+      color="cyan"
+      radius="md"
+      size="sm"
+      variant="filled"
+      style={{
+        float: 'left'
+      }}
```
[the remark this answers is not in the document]
sarahmcdougall: Good point, you're right it doesn't do anything and I forgot to remove it 😶

mgramigna (approved):

PR #29, opened 2023-04-17
Prism Highlighting for CQL for Library Resources

natjoe4:
# Summary
Populated the cql tab for Library resources. Added new CQL language to Prism for parsing and highlighting

## New behavior
Library resources with cql content will now have an enabled CQL tab in the resource viewer, populated with highlighted cql

## Code changes
- Added `react-prism-renderer` as a dependency
- Added new `Prism` component to `[resourceType]/[id]` in the cql tab and enabled the tab when appropriate
- Added regex for CQL highlighting to our local Prism version
- Updated .gitignore to remove some common files we don't want committed

# Testing guidance
 - Populate the backend server with resources using the `db:loadBundle` script, as well as `POST`ing resources. Make sure to include at least one Measure, one Library with cql content, and one Library without cql content.
 - Fire up the server and frontend with `npm run start:all`
 - Manipulate the url to view the resource page you wish to see. Make sure that when cql content is available on the resource, the cql tab is enabled and contains the highlighted cql. Make sure the tab is disabled when the resource is not a Library or when no cql content is available.


frontend/pages/[resourceType]/[id].tsx
```diff
@@ -3,13 +3,31 @@ import { Divider, Group, Stack, Tabs, Text } from '@mantine/core';
...
  */
 export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  const [decodedCql, setDecodedCql] = useState<string | null>(null);
+
+  // Overwrite Prism with our custom Prism that includes CQL as a language
+  useEffect(() => {
+    (window as any).Prism = PrismRenderer;
+  }, []);
+
+  useEffect(() => {
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:22]: In conjunction with my above comment and some of my recommendations regarding derived values in that [style guide](https://gist.github.com/mgramigna/bdc933fae2150216681c28d1f0f7aac7#hooks) that I made, I don't think `useEffect` is appropriate here since A) the content in the dependency array will never change and B) we can just derive the value in the comopnent itself.

If you want, you could memoize it instead so that, if other situations in this component cause a re-render, we don't re-compute the CQL:

```ts
const decodedCql = useMemo(() => {
  if (jsonData.resourceType === 'Measure') return null;
  const encodedCql = jsonData.content?.find(e => e.contentType === 'text/cql')?.data;
  return encodedCql ? Buffer.from(encodedCql, 'base64').toString() : null;
}, [jsonData]);
```

It's just an array lookup on something that will have like max 3 entries, so it's prob not that expensive. Although maybe on occasion the string could be a lot of bytes. idk

frontend/pages/[resourceType]/[id].tsx
```diff
@@ -3,13 +3,31 @@ import { Divider, Group, Stack, Tabs, Text } from '@mantine/core';
...
  * Component which displays the JSON/ELM/CQL/narrative content of an individual resource using
  * Mantine tabs
  * @returns JSON/ELM/CQL/narrative content of the individual resource in a Prism component
  */
 export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  const [decodedCql, setDecodedCql] = useState<string | null>(null);
+
+  // Overwrite Prism with our custom Prism that includes CQL as a language
+  useEffect(() => {
+    (window as any).Prism = PrismRenderer;
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/[id].tsx:19]: If you go with my comment in the helper, you'll have to set the `.languages.cql` property in here instead

mgramigna (approved):

frontend/prismLanguages/prismCQL.ts
```diff
@@ -0,0 +1,24 @@
+import { Prism } from 'prism-react-renderer';
```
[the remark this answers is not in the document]
natjoe4: Yeah I was a bit torn on where to put this. I agree that a whole directory is unnecessary if we're not gonna add additional languages

frontend/pages/[resourceType]/[id].tsx
Comment (natjoe4) [reply frontend/pages/[resourceType]/[id].tsx:22]: Always forget about useMemo. Good tip!

mgramigna (approved):

PR #30, opened 2023-04-19
Added view of resource narrative if it exists on resources

hossenlopp:
# Summary

Adds a tab to the resource page for the narrative if it exists.

## New behavior

Shows the narrative by pulling the HTML from the narrative right into the view.

## Code changes

New dependency `html-react-parser` to parse HTML string into react components. In `[id].tsx` add the tab and display of narrative. `global.css` add some override of default `th` elements to make it look better.

# Testing guidance

Load up resources into measure repository. In ecqm-content-r4-2021 the Library FHIR-ModelInfo has a narrative (i.e. `text`), ensure it shows up in the page for the resource. For other resources that do not have the `text` narrative field, ensure the "Narrative" tab does not appear.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -21,6 +21,7 @@
     "@mantine/notifications": "^6.0.6",
     "@mantine/prism": "^6.0.7",
     "@types/fhir": "^0.0.36",
+    "html-react-parser": "^3.0.15",
     "next": "13.2.4",
     "prism-react-renderer": "^1.3.5",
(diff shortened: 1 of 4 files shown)


sarahmcdougall (approved):
Great work on this! Good call on adding the custom styling to `globals.css` and having the tab only appear if the narrative is available.

elsaperelli (approved):
Nice!

PR #31, opened 2023-05-03
Replace <resourceType>/<id> links with info cards

sarahmcdougall:
# Summary
Replaces `<resourceType>/<id>` links with info cards for the resources available on the server. Each info card links to `[resourceType]/[id].tsx`.

## New Behavior
When browsing the available resources on the measure repository service, the user can browse info cards containing information about the resource beyond just its resource type and id. Clicking on a resource card opens it with more information, and a “view” button. Clicking the “view” button on a resource card will bring the user to the next page with the JSON/CQL/narrative content.

## Code Changes
* `ResourceInfoCard.tsx` - styling for “info cards” containing information about each resource. This includes `id`, `identifier`, `url`, `version`, `name`, and `status`. These fields are only included on the card if they are defined
* `[resourceType].tsx` - changes to server side props to get all the info needed for the info card. Update logic that loops over all the resources, replacing the existing `Link`/`Button` code with the `ResourceInfoCard`s
* `fhir.ts` - New interface for capturing all the resource info card information

# Testing Guidance
* Upload resources to the measure repository service - can use the `db:loadBundle` script to accomplish this
* Once the database is populated, use the resource list on the left-hand side to select a resource type —> this will bring you to `[resourceType].tsx`
* Select a resource info card by clicking on it 
* Inspect the style of the resource info card — sizing, color scheme, border color upon selection, order of the different fields, etc.
* Check that "viewing" a resource via the "view" button brings the user to appropriate page with the JSON/CQL/narrative content
* Double-check the typing of the `ResourceInfo` interface

Open Questions and Areas for Feedback:
* Thoughts on the card format? Should we have all the info appear on each card, or should we keep the current functionality, which only shows fields like name/version/status if the resource is selected? Trying to find a balance between how much info to show on the page at a time (as to not overwhelm the user) and aesthetics/scalability (since this is a demo app and I imagine we won’t have hundreds of resources on the server at a time)
    * If we want to display all the info at once, we could consider removing the “view” button entirely, and embed the link directly in the resource info card
* Currently, the identifier is only displayed on the card if both the `system` and `value` are defined. Open to thoughts on the best way to handle this, as `system` and `value` both have cardinality 0..1 (link to [spec](https://build.fhir.org/datatypes.html#Identifier) with more details)


frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <Text size="md" fw={700}>
+              {resourceInfo.id}
+            </Text>
+          </div>
+          {selected && (
+            <div>
+              {Object.entries(resourceInfo).map(([key, val]) => {
+                return (
+                  val && (
+                    <Text size="sm" color="dimmed">
```
mgramigna, line 44: These text nodes should have a unique key to avoid react rendering warnings:

image

frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+import React from 'react';
+import { ResourceInfo } from '@/util/types/fhir';
+import { ArrowNarrowRight } from 'tabler-icons-react';
+
+export interface ResourceInfoCardProps {
+  resourceInfo: ResourceInfo;
+  selected?: boolean;
+}
+
+export default function PatientInfoCard({ resourceInfo, selected }: ResourceInfoCardProps) {
```
mgramigna, line 12:
```suggestion
export default function ResourceInfoCard({ resourceInfo, selected }: ResourceInfoCardProps) {
```

?

frontend/pages/[resourceType].tsx
```diff
@@ -67,8 +49,24 @@ export default function ResourceList({ ids, resourceType }: InferGetServerSidePr
...
+                          setSelectedResource(res.id);
+                        }}
+                      >
+                        <ResourceInfoCard resourceInfo={res} selected={selectedResource === res.id} />
+                      </div>
+                    )
+                  );
+                })}
+              </Stack>
+            </div>
```
Comment (mgramigna) [inline frontend/pages/[resourceType].tsx:69]: Not related to your code but there a lowercase `text` node below this that's almost def supposed to be a maintine `<Text>` node instead

frontend/pages/[resourceType].tsx
```diff
@@ -104,8 +102,24 @@ export const getServerSideProps: GetServerSideProps<{
     // Measure Repository should not provide a bundle without an entry
     throw new Error('Measure Repository bundle has no entry.');
   }
-  const ids = bundle.entry.map(entry => entry.resource?.id);
+  const resources = bundle.entry as fhir4.BundleEntry<FhirArtifact>[];
```
Comment (natjoe4) [inline frontend/pages/[resourceType].tsx:105]: We can avoid double casting the bundle by casting it higher up to `fhir4.Bundle<FhirArtifact>`.

frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <div>
+            <Text size="md" fw={700}>
+              {resourceInfo.id}
+            </Text>
+          </div>
+          {selected && (
+            <div>
+              {Object.entries(resourceInfo).map(([key, val]) => {
+                return (
+                  val && (
```
mgramigna, line 43: Is there a reason you're doing this? Will `val` ever be nullish?

frontend/pages/[resourceType].tsx
```diff
@@ -67,8 +49,24 @@ export default function ResourceList({ ids, resourceType }: InferGetServerSidePr
...
         >
-          {idItems.length > 0 ? ( //if items exist
-            <ul>{idItems}</ul>
+          {resourceInfo.length > 0 ? ( // if items exist
+            <div>
+              <Stack style={{ marginTop: '10px', marginBottom: '10px', marginLeft: '50px', marginRight: '50px' }}>
+                {resourceInfo.map(res => {
+                  return (
+                    res && (
+                      <div
```
Comment (mgramigna) [inline frontend/pages/[resourceType].tsx:58]: same here re: These text nodes should have a unique key to avoid react rendering warnings:

frontend/pages/[resourceType].tsx
```diff
@@ -67,8 +49,24 @@ export default function ResourceList({ ids, resourceType }: InferGetServerSidePr
...
           }}
         >
-          {idItems.length > 0 ? ( //if items exist
-            <ul>{idItems}</ul>
+          {resourceInfo.length > 0 ? ( // if items exist
+            <div>
+              <Stack style={{ marginTop: '10px', marginBottom: '10px', marginLeft: '50px', marginRight: '50px' }}>
+                {resourceInfo.map(res => {
+                  return (
+                    res && (
```
Comment (mgramigna) [inline frontend/pages/[resourceType].tsx:57]: Same here re: why we need a conditional render, this should always be defined?

frontend/pages/[resourceType].tsx
Comment (natjoe4) [inline frontend/pages/[resourceType].tsx:69]: Wondering if it might be better to make this div scrollable, instead of the whole window. That way we could always have access to the back button. Thoughts?

frontend/pages/[resourceType].tsx
```diff
@@ -104,8 +102,24 @@ export const getServerSideProps: GetServerSideProps<{
     // Measure Repository should not provide a bundle without an entry
     throw new Error('Measure Repository bundle has no entry.');
   }
-  const ids = bundle.entry.map(entry => entry.resource?.id);
+  const resources = bundle.entry as fhir4.BundleEntry<FhirArtifact>[];
+  const resourceInfoArray = resources.reduce((acc: ResourceInfo[], entry) => {
+    if (entry.resource && entry.resource.id) {
+      const identifier = entry.resource.identifier?.[0];
+      const resourceInfo: ResourceInfo = {
```
Comment (mgramigna) [inline frontend/pages/[resourceType].tsx:109]: If any of these values (like `name`) end up as nullish, it can't be serialized as server side props as JSON, you'll get this error:

image

Should probably default values to null or omit them from the object entirely if they're undefined

frontend/util/types/fhir.ts
```diff
@@ -1,3 +1,17 @@
 // type representing the resource types that are relevant to the Measure Repository Service
 export type FhirArtifact = fhir4.Measure | fhir4.Library;
 export type ArtifactResourceType = FhirArtifact['resourceType'];
+
+/**
+ * Interface containing fields pulled from the resource JSON that are used
+ * to populate the Resource Info Card.
+ */
+export interface ResourceInfo {
```
mgramigna, line 9: Overall I like the approach of specifying what info we'll render on the actual page to limit the amount of JSON being sent as props (i.e. instead of sending over the entire list of resources or search set bundle)

frontend/pages/[resourceType].tsx
```diff
@@ -67,8 +49,24 @@ export default function ResourceList({ ids, resourceType }: InferGetServerSidePr
...
-            <ul>{idItems}</ul>
+          {resourceInfo.length > 0 ? ( // if items exist
+            <div>
+              <Stack style={{ marginTop: '10px', marginBottom: '10px', marginLeft: '50px', marginRight: '50px' }}>
+                {resourceInfo.map(res => {
+                  return (
+                    res && (
+                      <div
+                        onClick={() => {
+                          setSelectedResource(res.id);
```
Comment (natjoe4) [inline frontend/pages/[resourceType].tsx:60]: Can we add a check here that sees if the selected resource is the one clicked on, and sets `selectedResource` to null if so? That way we could collapse resources with a second click

frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+          style.border = `${rem(1)} solid ${theme.colors.blue[3]}`;
+        }
+        return style;
+      }}
+    >
+      <Grid align="center">
+        <Grid.Col span={10}>
+          <div>
+            <Text size="md" fw={700}>
+              {resourceInfo.id}
```
natjoe4, line 36: I know we talked a bit about what would be best displayed here. I'm wondering if `name` would be a better identifier when available? Also, do we want version displayed here if available? @hossenlopp, I feel like I remember you having opinions during the mockups meeting.

frontend/pages/[resourceType].tsx
Comment (mgramigna) [inline frontend/pages/[resourceType].tsx:105]: Not related to your code but there's a double equal string comparison above. Can you change that to `!==`?

natjoe4 (changes requested):
This looks awesome. Great work! Just a few smaller changes and design suggestions that are open to discussion :)

frontend/components/ResourceInfoCard.tsx
mgramigna, line 12: I have a lot of UI/UX paradigms/thoughts on this component. I'll list them out here, but let me know if you want to talk about any of them more:

1. I think there's no need to have a "collapsed" version that gets expanded later, I think the info should just be expanded always
2. Can we style the "View" button a little differently? It takes up a lot of space in the middle of the card. What if it was just some type of "expand icon" button in the top right or something, indicating that it will take you to a new page
3. This is the biggest one. It's very common when you start out rendering structured data to default to the approach of rendering `Key: Value`, as it would appear on an object.

We have a lot more freedom in a frontend app for how we render data. Take a look at the following example. Is my version intuitive enough that a user looking at it might know what each of these pieces of data is intuitively without needing a key value rendering? Maybe. I think we should push for less labels somehow if possible.

image

image

mgramigna (changes requested):

frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+            </Text>
+          </div>
+          {selected && (
+            <div>
+              {Object.entries(resourceInfo).map(([key, val]) => {
+                return (
+                  val && (
+                    <Text size="sm" color="dimmed">
+                      {`${key.charAt(0).toUpperCase() + key.slice(1)}: ${val}`}
```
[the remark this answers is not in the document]
mgramigna: See my below UI comments before addressing this

frontend/pages/[resourceType].tsx
Comment (mgramigna) [reply frontend/pages/[resourceType].tsx:109]: @natjoe4 already identified this in a comment while I was reviewing lol

frontend/pages/[resourceType].tsx
Comment (sarahmcdougall) [reply frontend/pages/[resourceType].tsx:57]: Good catch - originally the resource info array was constructed using a `.map` instead of a `.reduce` but I forgot to remove the conditional render after changing that logic

frontend/pages/[resourceType].tsx
```diff
@@ -64,17 +45,26 @@ export default function ResourceList({ ids, resourceType }: InferGetServerSidePr
...
+            </Stack>
           ) : (
-            <text>
+            <Text>
               No <i>{`${resourceType}`}</i> resources available
-            </text>
+            </Text>
           )}
-        </div>
+        </ScrollArea>
```
Comment (natjoe4) [inline frontend/pages/[resourceType].tsx:67]: This `ScrollArea` still extends off the page. Annoying component to work with. I've messed with it for a while now and can't find a way to make the resizing all work together smoothly. Would be down to chat about it if you're interested.

frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+          ) : (
+            <Text size="sm">{resourceInfo.status && <b>Status: {resourceInfo.status}</b>}</Text>
+          )}
+          {resourceInfo.url && (
+            <Text size="sm">
+              {resourceInfo.url}
+              {resourceInfo.version && `|${resourceInfo.version}`}
+            </Text>
+          )}
+          {resourceInfo.identifier && <Text size="sm">{`Identifier: ${resourceInfo.identifier}`}</Text>}
```
natjoe4, line 48: Love the new layout! I think the Identifier line is the only one that still looks a little off. I wonder if even just italicizing the word `Identifier` would be an improvement, or doing something more to signify that it's not part of the value

frontend/components/ResourceInfoCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+export default function ResourceInfoCard({ resourceInfo }: ResourceInfoCardProps) {
+  return (
+    <Paper
+      shadow="sm"
+      p="md"
+      sx={theme => {
+        const style: Sx = {
+          ':hover': {
+            cursor: 'pointer',
+            backgroundColor: theme.colors.gray[0]
```
natjoe4, line 20: Right now, we don't do anything when a resource is clicked. I'm not sure if it's more appropriate to remove this hover highlighting, or just make the whole resource card a link to the JSON content.

natjoe4 (changes requested):
This looks great. A few more small comments/conversation starters

mgramigna (commented):
I don't have anything in addition to what Nat pointed out

mgramigna (approved):

frontend/components/ResourceInfoCard.tsx
sarahmcdougall: I like the idea of italicizing `Identifier`. Here's what it looks like:
Screenshot 2023-05-09 at 9 01 27 AM

Let me know what you think - having trouble figuring out the best way to display this (or if we should even display it in the first place). I think this might be good for now and we can make adjustments later depending on the artifacts that we end up creating

natjoe4 (approved):
This looks awesome. Nice job! I think we could probably continue to improve the scrollarea, but if we want to tackle that I think it should be a new task. Approved!

PR #32, opened 2023-05-08
Artifact search page

elsaperelli:
# Summary
Adds a search page for all searchable content of artifacts that are currently supported by the measure repository service.

## New behavior
Users now have the ability to search for Measures and Libraries through the search parameters that are currently supported. 

## Code changes
- `ResourceButtons.tsx` - I moved the code that creates an array of resource buttons into a component since it is now being reused.
- `ResourceCounts.tsx` - Added a search button below the resource counts buttons array that links to a search page that has tabs for each artifact's search page.
- `SearchComponent.tsx` - A search input/preview component that contains inputs for each supported search parameter for a given artifact. Also builds and displays the request. Even though date search parameters are not yet supported, their inputs are handled for when they are eventually supported.
- `package.json` / `package-lock.json` - Upgraded mantine and added mantine/dates so I can handle date search inputs.
- `pages/[resourceType].tsx` - Added a search page to the resourceType page in the upper right corner. Replaced resource buttons code with new component.
- `pages/[resourceType]/search-result.tsx` - Search result page that displays the returned resources as resource buttons.
- `pages/[resourceType]/search.tsx` - A resourceType specific search page.
- `pages/search.tsx` - Overall search page for all resource types.
- `util/searchParams.ts` - Hardcoded search parameters for each artifact that the measure repository service currently supports.

# Testing guidance
- `npm run check:all`
- `npm run start:all`
- Navigate to `localhost:3001` and try different search inputs for both Measures and Libraries. Do so by navigating to the search page through both ways: clicking the search button in the nav bar AND the search button in the resourceType page.


lmd59:
Still looking through the code, but looks great so far! A couple of ideas on the UI side (feel free to disagree!):
- Would it make sense to put the nav area search button at the top above the resources? I don't think we're ever going to have more than 3 resource types, so we're always going to have a bunch of whitespace between the resources and the search button if it stays at the bottom there.
- On the search page, the resource types essentially act as tab labels, but this isn't really clear until you initially click on one. (Kind of looks like labels for information that failed to load). Maybe we can auto-select the first resource type if none are otherwise selected (once selected, they have a really clear tab look).
- This is just an opinion, but I kind of expected the search page to be the same no matter if you go from navigation or from the resourceType page, i.e. you can still switch tabs to other resource types if you'd like and it would just auto-select the tab for the resourceType page you navigated from. This might be a little bit more of a consistent user experience, but could definitely go either way. More so, I wonder if it would make the code a little simpler?
- I get this error for searching on an id that doesn't exist. It seems like it should fail a bit more gracefully:
Screen Shot 2023-05-08 at 3 00 33 PM
  - This also seems to happen for id's that *do* exist?
- The request preview box looks like an input field because it's similar to the rest of the input fields on the page. This would make sense if the request preview were editable, but it doesn't appear to be editable. As is, I would suggest losing the border box, and maybe making the text a bit gray or similar to show that the request preview is locked in. An editable preview would definitely be cool in the future, though!
- As the window loses a bit of width, some of the elements go a bit wonky. Below is a window that's about 2/3rds width of my laptop screen (I would say a reasonable width to be looking at), and the "search" button is starting to get cutoff when there seems like plenty of space for it to remain whole. As the "Request Preview" label goes to two lines, it doesn't align well with the content to the right anymore. Clear and search buttons get more cutoff as the width of the window shrinks.
Screen Shot 2023-05-08 at 3 05 13 PM
- Colors: red is generally reserved for destructive actions or errors, which I'm not sure clear really fits into (maybe a bit borderline so a little bit of red rather than the whole button?). I think you could get away with some more neutral colors on these buttons - could add clear and search icons to the buttons to help differentiate them for some more pizzaz.

frontend/components/SearchComponent.tsx
```diff
@@ -0,0 +1,135 @@
...
+            label={si.name}
+            value={si.value}
+            onChange={event => changeSearchInputs(si.name, event.currentTarget.value)}
+          />
+        </Grid.Col>
+      );
+    } else {
+      return (
+        <Grid.Col span={6} key={si.name}>
+          <DateInput label={si.name} value={si.date} onChange={event => changeSearchInputs(si.name, si.value, event)} />
```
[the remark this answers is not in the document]
elsaperelli: Yes! So I thought we supported date so I did this and then I realized we don't haha so I figured we could keep it when we do support date (I am pretty sure we eventually will) or I can just take it out.

frontend/components/SearchComponent.tsx
```diff
@@ -0,0 +1,185 @@
...
+    const requestParams: Parameter[] = [];
+    let urlAndVersion: string;
+    searchInputs.forEach(si => {
+      if (si.name === 'url') {
+        urlAndVersion = encodeURIComponent(si.value);
+      } else if (si.name === 'version') {
+        if (urlAndVersion !== '') {
+          si.value !== ''
+            ? requestParams.push(
+                { name: 'url', description: si.description, value: encodeURIComponent(urlAndVersion) },
```
lmd59, line 99:
```suggestion
                { name: 'url', description: si.description, value: urlAndVersion },
```
This actually re-encodes the `urlAndVersion` already encoded above

lmd59 (commented):
Great updates!
Other general comments:

- If you've made any progress on the individual resource search buttons going to the tabbed search (previous bullet 3), would be happy to chat about that more! I think that approach might also fix the back button issue we have on the tabbed page as well since (with it always defaulting to Measure) because we would have the resourceType as a parameter.
- I honestly don't think the search parameter descriptions are super helpful as is. I know that they come from the spec, but the spec is not really adding any information with these descriptions. I don't know if it's worth hand tweaking to be something useful...? That might not make sense to hardcode up though.
- 100% opinion, but the clear/search buttons still seem a bit cluttered to me. I might consider using more neutral colors, but also I kind of wouldn't mind @mgramigna's opinion on it since he's planning on reviewing and a bit more familiar with good design principles.

mgramigna:
> 100% opinion, but the clear/search buttons still seem a bit cluttered to me. I might consider using more neutral colors, but also I kind of wouldn't mind @mgramigna's opinion on it since he's planning on reviewing and a bit more familiar with good design principles.

I would delete the clear button entirely. I don't think it will be super common to fill out all the fields and then need to clear it ever. Realistically we'll probably only ever be searching by two fields max and they will rarely change.

For the search button, I might make it blue instead of green but I don't care too much about that one

frontend/components/SearchComponent.tsx
```diff
@@ -0,0 +1,185 @@
...
+   */
+  const requestPreview = () => {
+    let requestPreview = '';
+    const requestParams: Parameter[] = [];
+    let urlAndVersion: string;
+    searchInputs.forEach(si => {
+      if (si.name === 'url') {
+        urlAndVersion = encodeURIComponent(si.value);
+      } else if (si.name === 'version') {
+        if (urlAndVersion !== '') {
```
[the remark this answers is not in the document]
elsaperelli: Good catch!

frontend/pages/[resourceType]/search-result.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  // Cast to ArtifactResourceType since we know the server should only support resourceType that matches
+  const checkedResourceType = resourceType as ArtifactResourceType;
+
+  const url = context.resolvedUrl.split('search-result')[1];
+  const res = await fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/${resourceType}${url}`);
+  const json = await res.json();
+
+  if (json.resourceType === 'OperationOutcome') {
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/search-result.tsx:74]: I'm curious. Does this ever happen? I would assume that we get an `OperationOutcome` if the server responds with a `4XX` or `5XX`. I would expect the call to `await fetch` or `await .json()` to throw an error when that happens meaning that the code would abort before reaching here. Not sure though

frontend/pages/[resourceType]/search.tsx
```diff
@@ -0,0 +1,40 @@
...
+          </Group>
+        </div>
+        <Divider my="sm" style={{ paddingBottom: '6px' }} />
+        <SearchComponent resourceType={resourceType} />
+      </Stack>
+    </div>
+  );
+}
+
+export const getServerSideProps: GetServerSideProps<{
```
Comment (mgramigna) [inline frontend/pages/[resourceType]/search.tsx:26]: Does this need to be a `getServerSideProps` at all? That should be reserved for when we are doing something that requires backend communication. It looks like this is just checking request.query for a resource type, so it might make more sense to just do that on the client and render an error somewhere if it happens? If a user goes to `/Patient/search` or something then shame on them lol

mgramigna (changes requested):

mgramigna:
> This is just an opinion, but I kind of expected the search page to be the same no matter if you go from navigation or from the resourceType page, i.e. you can still switch tabs to other resource types if you'd like and it would just auto-select the tab for the resourceType page you navigated from. This might be a little bit more of a consistent user experience, but could definitely go either way. More so, I wonder if it would make the code a little simpler?

I agree with this comment, happy to talk through implementation

frontend/pages/[resourceType]/search-result.tsx
Comment (mgramigna) [reply frontend/pages/[resourceType]/search-result.tsx:74]: Gotcha okay cool. I thought `fetch` would throw an error if it gets `400` response code but turns out that isn't true

mgramigna (approved):

lmd59:
This all looks great! Thanks for making updates. A quick suggestion to look into, but if it presents headaches, we can ignore for now...
With the new ability to use resourceType parameters on the search inputs page to select a tab, it would be nice if we could do the reverse of sorts. I.e. when the user selects the library tab, the resourceType gets added to the query url to make the url: `http://localhost:3001/search?resourceType=Library`, and when the user selects the measure tab, we update the url to `http://localhost:3001/search?resourceType=Measure`. This would make the page consistent for back button and refresh.

lmd59 (approved):
Nice- awesome feature!

PR #33, opened 2023-05-08
Show resources, operations, and search parameters that are part of th…

lmd59:
…e capability statement

# Summary
List resource types, operations, and search parameters that are supported, according to the capability statement. Suggested UI: tabular.

## New behavior
Shows the capability statement information on the home page under the capabilities section.

## Code changes
The main `index.tsx` uses getServerSideProps to populate a capability statement property, which is then used to create a capability statement table in a new Capability Statement titled section. Additionally, small changes to colons to make consistent across page.

# Testing guidance
- `npm run check:all`
- `npm run start:all`
- Navigate to http://localhost:3001/ and view new Capability Statement section on the homepage
- Also check behavior for backend not started with `npm run start:frontend` and navigate to http://localhost:3000/

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -41,7 +41,13 @@ export default function Home({ capabilityStatement }: InferGetServerSidePropsTyp
                 <th>Search Parameters</th>
               </tr>
             </thead>
-            <tbody>{capabilityStatement.rest?.[0]?.resource?.map(r => CapabilityRow(r))}</tbody>
+            <tbody>
+              {capabilityStatement.rest?.[0]?.resource
+                ?.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


natjoe4 (changes requested):
This looks great to me! Just a couple small changes. On suggestion, do we want to sort operations and search params alphabetically? We don't have many, so things are easy enough to find either way, but could be a nice touch, especially if we plan to add more search params

natjoe4 (approved):
Looks great!

elsaperelli (changes requested):
Nice! My only suggestion is now that the resources and search parameters are alphabetized in the capability statement section, should we alphabetize the resources in the navbar so it matches? Aka Library then Measure.

elsaperelli (approved):

PR #34, opened 2023-05-09
Added id to core search args

natjoe4:
# Summary
Id was left out of the core search args. This adds it.

## New behavior
We can now do queries including `id=resourceId`. Previously, we could only search by id in the url args like `localhost:3000/4_0_1/Measure/resourceId`

## Code changes
added id to core search args

# Testing guidance
- `npm run check`
- Test search functionality and make sure queries including and excluding `id` work as expected.


elsaperelli (approved):
Woo!

PR #39, opened 2023-06-08
Renders ELM for selected Library resource

BriannaVH:
# Summary 
Renders ELM for selected Library resource

## New behavior
Users can now see the rendered ELM for a selected Library resource upon clicking the ELM tab.

## Code changes
`app/src/pages/[resourceType]/[id].tsx` - I enabled the ELM tab so that users can actually click on it when it appears. I added a new Prism component in the ELM tab and make the ELM tab appear/disappear whenever appropriate.

# Testing guidance
`npm run check:all`
`npm run start:all`
Navigate to http://localhost:3001/ and select both Measure and Library packages. Check that the ELM tab appears whenever the CQL tab appears, and likewise disappears if the CQL tab isn't present. The ELM tab should not appear for any Measures. Additionally, check that the ELM tab is enabled whenever it appears on the screen and that it appropriately renders ELM whenever it is selected by the user.

### DIFF
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -21,8 +21,8 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     /* eslint-enable @typescript-eslint/no-explicit-any */
   }, []);
 
-  const decodedCql = decode('text/cql', jsonData);
-  const decodedElm = decode('application/elm+json', jsonData);
+  const decodedCql = Decode('text/cql', jsonData);
+  const decodedElm = Decode('application/elm+json', jsonData);
 
   return (
     <div>
@@ -76,17 +76,16 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
 }
 
 /**
- *
- * Function which extracts a specific language from the fhir artifact and then returns the decoded
- * version of that language
+ * Function which extracts specified content from JSON data and then returns the decoded
+ * version of that content (in this case the ELM/CQL)
  * @returns The decoded version of the ELM/CQL content
  */
-function decode(link: String, jsonData: FhirArtifact) {
+function Decode(link: string, jsonData: FhirArtifact) {
   const decodedLanguage = useMemo(() => {
     if (jsonData.resourceType === 'Measure') return null;
     const encodedLanguage = (jsonData as fhir4.Library).content?.find(e => e.contentType === link)?.data;
     return encodedLanguage ? Buffer.from(encodedLanguage, 'base64').toString() : null;
-  }, [jsonData]);
+  }, [jsonData, link]);
   return decodedLanguage;
 }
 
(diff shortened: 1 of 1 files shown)


app/src/pages/[resourceType]/[id].tsx
```diff
@@ -61,6 +65,15 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
               </Prism>
             </Tabs.Panel>
           )}
+          {decodedELM != null && (
+            <Tabs.Panel value="elm" pt="xs">
+              {/* eslint-disable  @typescript-eslint/no-explicit-any */}
+              <Prism language={'elm' as any} colorScheme="light">
```
Comment (hossenlopp) [inline app/src/pages/[resourceType]/[id].tsx:71]:
```suggestion
              <Prism language="json" colorScheme="light">
```
The ELM is JSON and highlighting is supported in Prism so you can just use "json" for the language. Similar to how it is used for the resource JSON just above. Only difference here is that we already have the JSON as a string.

hossenlopp (changes requested):
The ELM JSON could be highlighted with a quick change. Other than that, it looks good!

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -61,6 +65,15 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
               </Prism>
             </Tabs.Panel>
           )}
+          {decodedELM != null && (
+            <Tabs.Panel value="elm" pt="xs">
+              {/* eslint-disable  @typescript-eslint/no-explicit-any */}
+              <Prism language="json" colorScheme="light">
+                {/* eslint-enable  @typescript-eslint/no-explicit-any */}
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:72]: Now that we are properly using `json` as the language, we can get rid of these comments that disable the lint rules for no explicit any.

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -73,6 +75,20 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+ * version of that content (in this case the ELM/CQL)
+ * @returns The decoded version of the ELM/CQL content
+ */
+function Decode(link: string, jsonData: FhirArtifact) {
+  const decodedLanguage = useMemo(() => {
+    if (jsonData.resourceType === 'Measure') return null;
+    const encodedLanguage = (jsonData as fhir4.Library).content?.find(e => e.contentType === link)?.data;
+    return encodedLanguage ? Buffer.from(encodedLanguage, 'base64').toString() : null;
+  }, [jsonData, link]);
+  return decodedLanguage;
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:89]: Suggest moving the `useMemo` part back to the component and just have this function do the extraction.

mgramigna (changes requested):

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -73,6 +80,17 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
   );
 }
 
+/**
+ * Function which extracts specified content from JSON data and then returns the decoded
+ * version of that content (in this case the ELM/CQL)
+ * @returns The decoded version of the ELM/CQL content
+ */
+function Decode(link: string, jsonData: FhirArtifact) {
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:88]: Last suggestion is to make this function lowercase `decode`. Uppercase/PascalCase should be reserved for react components and classes

mgramigna (approved):
Great work!

hossenlopp (approved):
This PR ended up being super clean!

PR #40, opened 2023-06-12
Draft existing artifacts

elsaperelli:
# Summary
This PR allows the user to pull an existing artifact (Measure or Library) from the measure repository service and clone it with a "draft" status instead of "active" status. This adds this new cloned artifact to the local database.

## New behavior
Before, the user could only create new artifacts from scratch and add it to the local database. Now the user can do so using an existing artifact on the measure repository service. The user can do this in the application in one of two ways: navigate to an existing artifact and click "Clone to Draft Artifact" OR go to the authoring page and under "Start From Existing Measure" the user can select a resource to clone.

## Code changes
- `app/src/pages/[resourceType]/[id].tsx` - new button on a specified measure repository service resource page to clone the artifact to a draft.
- `app/src/pages/api/trpc/[trpc].ts` - increased the NextJS body parser limit (errored out when I tried to clone the MATGlobalCommonFunctionsFHIR4 Library).
- `app/src/pages/authoring/[resourceType]/[id].tsx` - fills the inputs with the correct url and identifier when an artifact is cloned.
- `app/src/pages/authoring/index.tsx` - Added a dropdown on the authoring page for a user to clone an existing artifact.
- `app/src/server/trpc/routers/service.ts` - added two new endpoints: `getArtifactsByResource` (returns all of the artifacts of a certain resourceType) and `convertArtifactById` (takes in a resourceType and an id and calls `createDraft` on that artifact).

# Testing guidance
- `npm run check:all`
- Try cloning Library and Measure artifacts in the MRS through the two ways- directly from the artifact and through the dropdown on the authoring page. Things to look out for:
   - Make sure the number of Library and Measure artifacts increases in the Authoring page nav bar.
   - Go into studio 3T and make sure the artifacts are indeed in the local database once they are cloned.
   - Make sure the status of the cloned artifact is `draft` and not `active`.
   - Make sure that the `url` and `identifier` inputs are populated with the correct values once an artifact is cloned.
   - Make sure the success/error notifications work and are correct based on whether a resource was created or cloned.

### DIFF
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -63,6 +63,7 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
   const cloneResource = () => {
     const newClonedResource = jsonData;
     newClonedResource.id = uuidv4();
+    newClonedResource.status = 'draft';
     draftMutation.mutate({ resourceType, draft: newClonedResource });
   };
 
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -42,7 +42,7 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     onSuccess: data => {
       notifications.show({
         title: `Draft ${jsonData.resourceType} Created!`,
-        message: `Draft ${jsonData.resourceType} successfully created from ${jsonData.resourceType}/${jsonData.id}`,
+        message: `Draft ${jsonData.resourceType}/${jsonData.id} successfully created`,
         icon: <CircleCheck />,
         color: 'green'
       });
@@ -52,7 +52,7 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     onError: e => {
       notifications.show({
         title: `Draft ${jsonData.resourceType} Creation Failed!`,
-        message: `Attempt to create draft ${jsonData.resourceType} from ${jsonData.resourceType}/${jsonData.id} failed with message: ${e.message}`,
+        message: `Attempt to create draft of ${jsonData.resourceType}/${jsonData.id} failed with message: ${e.message}`,
         icon: <AlertCircle />,
         color: 'red'
       });
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -11,6 +11,7 @@ import { AlertCircle, CircleCheck } from 'tabler-icons-react';
 import { notifications } from '@mantine/notifications';
 import { trpc } from '../../util/trpc';
 import { useRouter } from 'next/router';
+
 /**
  * Component which displays the JSON/ELM/CQL/narrative content of an individual resource using
  * Mantine tabs
@@ -40,8 +41,8 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
   const draftMutation = trpc.draft.createDraft.useMutation({
     onSuccess: data => {
       notifications.show({
-        title: `${jsonData.resourceType} Cloned!`,
-        message: `${jsonData.resourceType} successfully cloned`,
+        title: `${jsonData.resourceType} Created!`,
+        message: `${jsonData.resourceType} successfully created from ${jsonData.resourceType}/${jsonData.id}`,
         icon: <CircleCheck />,
         color: 'green'
       });
@@ -50,8 +51,8 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     },
     onError: e => {
       notifications.show({
-        title: `${jsonData.resourceType} Clone Failed!`,
-        message: `Attempt to clone ${jsonData.resourceType} failed with message: ${e.message}`,
+        title: `${jsonData.resourceType} Creation Failed!`,
+        message: `Attempt to create ${jsonData.resourceType} from ${jsonData.resourceType}/${jsonData.id} failed with message: ${e.message}`,
         icon: <AlertCircle />,
         color: 'red'
       });
@@ -60,11 +61,11 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
 
   const router = useRouter();
 
-  const cloneResource = () => {
-    const newClonedResource = jsonData;
-    newClonedResource.id = uuidv4();
-    newClonedResource.status = 'draft';
-    draftMutation.mutate({ resourceType, draft: newClonedResource });
+  const createDraftOfArtifact = () => {
+    const draftOfArtifact = jsonData;
+    draftOfArtifact.id = uuidv4();
+    draftOfArtifact.status = 'draft';
+    draftMutation.mutate({ resourceType, draft: draftOfArtifact });
   };
 
   return (
@@ -75,8 +76,8 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
             <Text size="xl" color="gray">
               {jsonData.resourceType}/{jsonData.id}
             </Text>
-            <Button w={240} loading={draftMutation.isLoading} onClick={cloneResource}>
-              Clone to Draft Artifact
+            <Button w={240} loading={draftMutation.isLoading} onClick={createDraftOfArtifact}>
+              Create Draft of {jsonData.resourceType}
             </Button>
           </Group>
         </div>
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -64,6 +64,7 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     const draftOfArtifact = jsonData;
     draftOfArtifact.id = uuidv4();
     draftOfArtifact.status = 'draft';
+    delete draftOfArtifact.version;
     draftMutation.mutate({ resourceType, draft: draftOfArtifact });
   };
 
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -6,11 +6,11 @@ import { useEffect, useMemo } from 'react';
 import CQLRegex from '../../util/prismCQL';
 import { Prism as PrismRenderer } from 'prism-react-renderer';
 import parse from 'html-react-parser';
-import { v4 as uuidv4 } from 'uuid';
 import { AlertCircle, CircleCheck } from 'tabler-icons-react';
 import { notifications } from '@mantine/notifications';
 import { trpc } from '../../util/trpc';
 import { useRouter } from 'next/router';
+import { modifyResourceToDraft } from '@/util/modifyResourceFields';
 
 /**
  * Component which displays the JSON/ELM/CQL/narrative content of an individual resource using
@@ -61,10 +61,7 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
   });
 
   const createDraftOfArtifact = () => {
-    const draftOfArtifact = jsonData;
-    draftOfArtifact.id = uuidv4();
-    draftOfArtifact.status = 'draft';
-    delete draftOfArtifact.version;
+    const draftOfArtifact = modifyResourceToDraft(jsonData);
     draftMutation.mutate({ resourceType, draft: draftOfArtifact });
   };
 
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -41,8 +41,8 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
   const draftMutation = trpc.draft.createDraft.useMutation({
     onSuccess: data => {
       notifications.show({
-        title: `${jsonData.resourceType} Created!`,
-        message: `${jsonData.resourceType} successfully created from ${jsonData.resourceType}/${jsonData.id}`,
+        title: `Draft ${jsonData.resourceType} Created!`,
+        message: `Draft ${jsonData.resourceType} successfully created from ${jsonData.resourceType}/${jsonData.id}`,
         icon: <CircleCheck />,
         color: 'green'
       });
@@ -51,8 +51,8 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     },
     onError: e => {
       notifications.show({
-        title: `${jsonData.resourceType} Creation Failed!`,
-        message: `Attempt to create ${jsonData.resourceType} from ${jsonData.resourceType}/${jsonData.id} failed with message: ${e.message}`,
+        title: `Draft ${jsonData.resourceType} Creation Failed!`,
+        message: `Attempt to create draft ${jsonData.resourceType} from ${jsonData.resourceType}/${jsonData.id} failed with message: ${e.message}`,
         icon: <AlertCircle />,
         color: 'red'
       });
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -1,18 +1,24 @@
 import { Prism } from '@mantine/prism';
-import { Divider, Group, Space, Stack, Tabs, Text } from '@mantine/core';
+import { Button, Divider, Group, Space, Stack, Tabs, Text } from '@mantine/core';
 import { GetServerSideProps, InferGetServerSidePropsType } from 'next';
 import { FhirArtifact } from '@/util/types/fhir';
 import { useEffect, useMemo } from 'react';
 import CQLRegex from '../../util/prismCQL';
 import { Prism as PrismRenderer } from 'prism-react-renderer';
 import parse from 'html-react-parser';
-
+import { v4 as uuidv4 } from 'uuid';
+import { AlertCircle, CircleCheck } from 'tabler-icons-react';
+import { notifications } from '@mantine/notifications';
+import { trpc } from '../../util/trpc';
+import { useRouter } from 'next/router';
 /**
  * Component which displays the JSON/ELM/CQL/narrative content of an individual resource using
  * Mantine tabs
  * @returns JSON/ELM/CQL/narrative content of the individual resource in a Prism component
  */
 export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  const resourceType = jsonData.resourceType;
+
   // Overwrite Prism with our custom Prism that includes CQL as a language
   useEffect(() => {
     /* eslint-disable @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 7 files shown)


app/src/pages/[resourceType]/[id].tsx
```diff
@@ -27,14 +33,49 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+  });
+
+  const router = useRouter();
+
+  const cloneResource = () => {
+    const newClonedResource = jsonData;
+    newClonedResource.id = uuidv4();
+    newClonedResource.status = 'draft';
+    draftMutation.mutate({ resourceType, draft: newClonedResource });
+  };
```
Comment (BriannaVH) [inline app/src/pages/[resourceType]/[id].tsx:66]: I'm not sure if this what is happening, but would it be possible to only get the `cloneResource` after the `clone to draft artifact button` is clicked and is successful?

app/src/pages/authoring/index.tsx
```diff
@@ -16,28 +16,54 @@ const useStyles = createStyles(() => ({
...
+      ? `Attempt to clone ${resourceType} failed with message: ${errorMessage}`
+      : `Attempt to create ${resourceType} failed with message: ${errorMessage}`;
+    notifications.show({
+      title: title,
+      message: message,
+      icon: <AlertCircle />,
+      color: 'red'
+    });
+  };
+
```
BriannaVH, line 51: You also have a success/error notification pop up in `[app/src/pages/[resourceType]/[id].tsx]` on lines 39-57 and I was wondering if it would be helpful to use/modify these two methods: errorNotification() and successNotification() in that file as well to reduce code?

elsaperelli: Good suggestion! I think I will pull them out into a `util/draftArtifactNotifications` file if that makes sense.

app/src/pages/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx:66]: Do you mean like only see the JSON of the resource without those input buttons? Or something else?

app/src/pages/authoring/index.tsx
elsaperelli: Decided to keep this is way it is based on our conversation in dev hour.

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -27,14 +33,49 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
     <div>
       <Stack spacing="xs">
         <div>
-          <Group>
+          <Group position="apart">
             <Text size="xl" color="gray">
               {jsonData.resourceType}/{jsonData.id}
             </Text>
+            <Button w={240} loading={draftMutation.isLoading} onClick={cloneResource}>
+              Clone to Draft Artifact
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:77]: I think the nomenclature might get confusing here. [Clone](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#clone) is a separate operation than draft, which is very similar. Maybe instead of the word `clone` we can say something else.

* Create Draft of <some identifying string>
* Create Draft of <resource type>
* Something else?

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -16,11 +16,28 @@ export default function ResourceAuthoringPage() {
...
+      const cmsIdentifier = resource.identifier.find(
+        identifier => identifier.system === 'http://hl7.org/fhir/cqi/ecqm/Measure/Identifier/cms'
+      );
+      if (cmsIdentifier?.value) {
+        setIdentifier(cmsIdentifier.value);
+      } else if (resource.identifier[0].value) {
+        setIdentifier(resource.identifier[0].value);
+      }
+    }
+  }, [resource?.url, resource?.identifier]);
```
Comment (mgramigna) [inline app/src/pages/authoring/[resourceType]/[id].tsx:39]: Maybe let's just put `resource` as the dependency, assuming we will have more inputs in the future, and I never really like using `?` in useEffect dependencies

app/src/pages/authoring/index.tsx
```diff
@@ -65,6 +97,23 @@ export default function AuthoringPage() {
...
+          <Title order={3}>{`Start From an Existing ${resourceType}:`}</Title>
+          {artifacts ? (
+            <Select
+              label={`Select an existing ${resourceType} to clone`}
+              data={artifacts}
+              value={selectedArtifact}
+              onChange={setSelectedArtifact}
+            />
+          ) : (
+            <Text>No artifacts</Text>
```
mgramigna, line 109: Instead of rendering text, can we make use of the `isLoading` attribute of the tRPC query to either disable the dropdown, render a spinner for it instead, etc.?

app/src/pages/authoring/index.tsx
```diff
@@ -65,6 +97,23 @@ export default function AuthoringPage() {
...
+          )}
+          <Button
+            w={240}
+            loading={draftCloneMutation.isLoading}
+            onClick={cloneResource}
+            disabled={!selectedArtifact}
+          >{`Clone Draft ${resourceType}`}</Button>
```
mgramigna, line 116: No need to use string interpolation here:

```suggestion
          >Clone Draft {resourceType}</Button>
```

I see there are some other buttons from other code that do this too, wanna just update those as well?


But also see above comment about use of the word "clone"

mgramigna (changes requested):
A couple inline comments, plus the following:

From [the spec](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#draft):

> The draft operation supports creating a new draft version of an existing artifact in active status. This operation creates a new resource with the same contents as the existing artifact, but with a status of draft and no version. It is an error to create multiple drafts of the same artifact by URL.

We didn't really account for the last sentence when thinking about this task. Looks like you are only allowed to have one draft in the application for a given artifact URL? I think this will be a whole separate undertaking that we should probably just make a task for.

app/src/pages/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx:77]: Hmmm...I think I'll go with Create Draft of 😆

Comment (mgramigna) [reply app/src/pages/[resourceType]/[id].tsx:77]: lol GitHub escaped my comment. I edited it

Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx:77]: For the "Cloned!" and "Successfully cloned!" messages, can I just change them to "Created!" and "Successfully created" or should I distinguish them from the create draft from scratch somehow?

app/src/pages/[resourceType]/[id].tsx
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: Sorry, I was a bit confused on this piece of code earlier and I think it led me to mistakenly think something was wrong. I'm still a bit confused but everything looks good to me !

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -16,11 +16,28 @@ export default function ResourceAuthoringPage() {
 
   const ctx = trpc.useContext();
 
-  const resourceQuery = trpc.draft.getDraftById.useQuery({
+  const { data: resource } = trpc.draft.getDraftById.useQuery({
```
Comment (mgramigna) [inline app/src/pages/authoring/[resourceType]/[id].tsx:19]: this causes a trpc error when the page first loads, as `resourceType` will be undefined on first load. Maybe allow for the `getDraftbyId` input to have resourceType as optional, and if it's not provided we just return null or something

app/src/pages/authoring/index.tsx
```diff
@@ -1,4 +1,15 @@
-import { Button, Center, Paper, SegmentedControl, Stack, Title, createStyles } from '@mantine/core';
+import {
+  Button,
+  Center,
+  Paper,
+  SegmentedControl,
+  Select,
+  Stack,
+  Title,
+  Text,
```
mgramigna, line 9: Text is unused

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -89,7 +106,7 @@ export default function ResourceAuthoringPage() {
         <Grid.Col span={6}>
           <Paper withBorder>
             <Prism language="json" colorScheme="light">
-              {resourceQuery.data ? JSON.stringify(resourceQuery.data, null, 2) : ''}
+              {resource ? JSON.stringify(resource, null, 2) : ''}
```
Comment (mgramigna) [inline app/src/pages/authoring/[resourceType]/[id].tsx:109]: This is now huge for real resources. I think we might want to limit the height of the JSON view and make it scrollable, essentially allowing a user to scroll in the JSON while keeping the inputs on the left side where they are

mgramigna (changes requested):

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -88,9 +113,11 @@ export default function ResourceAuthoringPage() {
         </Grid.Col>
         <Grid.Col span={6}>
           <Paper withBorder>
-            <Prism language="json" colorScheme="light">
-              {resourceQuery.data ? JSON.stringify(resourceQuery.data, null, 2) : ''}
-            </Prism>
+            <div className={classes.json}>
+              <Prism language="json" colorScheme="light">
```
Comment (mgramigna) [inline app/src/pages/authoring/[resourceType]/[id].tsx:117]: It makes more sense here to use the [styles API](https://mantine.dev/others/prism/?t=styles-api) to customize the default scroll area container of the `Prism` rather than wrap it in a div. With this div, we need to scroll all the way to the bottom to get a horizontal scrollbar, and the copy button disappears. Something like this should work better:

```
              <Prism
                styles={{
                  scrollArea: {
                    height: 'calc(100vh - 150px)'
                  }
                 // ...
                }}
              >
```

And remove the custom `json` class from the parent

app/src/pages/authoring/index.tsx
```diff
@@ -65,6 +97,25 @@ export default function AuthoringPage() {
           <Button w={240} loading={draftMutation.isLoading} onClick={createResource}>
             {`Create New Draft ${resourceType}`}
           </Button>
+          <Title order={3}>{`Start From an Existing ${resourceType}:`}</Title>
```
mgramigna, line 100:
```suggestion
          <Title order={3}>Start From an Existing {resourceType}:</Title>
```

app/src/pages/authoring/index.tsx
```diff
@@ -65,6 +97,25 @@ export default function AuthoringPage() {
...
+          {artifacts ? (
```
mgramigna, line 101: I still think we should make use of the other attributes that we get back from a trpc `useQuery`. If there is an error or something, the user will just see a loading spinner with no info. A common pattern in trpc is something like this:

```tsx
const {
  data: artifacts,
  isLoading: artifactsLoading,
  error: artifactError
} = trpc.service.getArtifactsByResource.useQuery({ resourceType });


// ...

{artifactsLoading ? (
  <Loader />
) : artifactError ? (
  <div>{artifactError.message}</div>
) : artifacts ? (
  <Select
    label={`Select an existing ${resourceType} to create a draft from`}
    data={artifacts}
    value={selectedArtifact}
    onChange={setSelectedArtifact}
  />
) : (
  <div>An unknown error occurred fetching artifacts</div>
)}
```

You might want to do more styling than just a div for rendering the error message(s). The above is just to show the structure

mgramigna (changes requested):

app/src/pages/authoring/index.tsx
```diff
@@ -16,28 +16,54 @@ const useStyles = createStyles(() => ({
...
+  const [selectedArtifact, setSelectedArtifact] = useState<string | null>(null);
+
+  const { data: artifacts } = trpc.service.getArtifactsByResource.useQuery({ resourceType });
 
   const ctx = trpc.useContext();
   const { classes } = useStyles();
 
+  const successNotification = (data: { draftId: string }, createdFromArtifact: boolean) => {
+    const message = createdFromArtifact
+      ? `${resourceType} successfully created from ${resourceType}/${selectedArtifact}`
```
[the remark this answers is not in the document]
sarahmcdougall: Could also make the same change to the success notification in `app/src/pages/[resourceType]/[id].tsx`

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -29,14 +36,49 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+      notifications.show({
+        title: `${jsonData.resourceType} Creation Failed!`,
+        message: `Attempt to create ${jsonData.resourceType} from ${jsonData.resourceType}/${jsonData.id} failed with message: ${e.message}`,
+        icon: <AlertCircle />,
+        color: 'red'
+      });
+    }
+  });
+
+  const router = useRouter();
```
Comment (sarahmcdougall) [inline app/src/pages/[resourceType]/[id].tsx:62]: Nitpick but I would move this line to be above `draftMutation` since `draftMutation` pushes to the router but is defined before the router is

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -16,11 +24,28 @@ export default function ResourceAuthoringPage() {
 
   const ctx = trpc.useContext();
 
-  const resourceQuery = trpc.draft.getDraftById.useQuery({
+  const { data: resource } = trpc.draft.getDraftById.useQuery({
     id: id as string,
     resourceType: resourceType as ArtifactResourceType
   });
 
+  // useEffect to check if it has an identifier and url
```
Comment (sarahmcdougall) [inline app/src/pages/authoring/[resourceType]/[id].tsx:32]:
```suggestion
  // useEffect to check if the resource has an identifier and url defined
```

app/src/server/trpc/routers/service.ts
```diff
@@ -16,5 +20,33 @@ export const serviceRouter = router({
...
+  convertArtifactById: publicProcedure
+    .input(z.object({ resourceType: z.enum(['Measure', 'Library']), id: z.string() }))
+    .mutation(async ({ input }) => {
+      const draftRes = await fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/${input.resourceType}/${input.id}`);
+      const draftArtifact = (await draftRes.json()) as FhirArtifact;
+      draftArtifact.id = uuidv4();
+      draftArtifact.status = 'draft';
+
+      const draftCaller = draftRouter.createCaller({});
+      const res = await draftCaller.createDraft({ resourceType: input.resourceType, draft: draftArtifact });
```
sarahmcdougall, line 49: Still trying to wrap my head around `createCaller` since this is my first time seeing it used in context. I was reading through the [tRPC docs](https://trpc.io/docs/server/server-side-calls) on server-side calls, which says the following: 

> createCaller should not be used to call procedures from within other procedures. This creates overhead by (potentially) creating context again, executing all middlewares, and validating the input - all of which were already done by the current procedure.

Are we doing this here or not? Seems like `draftRouter` has a procedure `createDraft`, and then `serviceRouter` now has this procedure `getArtifactsByResource` that calls the `createDraft` procedure.

sarahmcdougall (changes requested):
Aside from the in-line comments, I saw the following in the spec:

> This operation creates a new resource with the same contents as the existing artifact, but with a status of draft and no version.

When I tested this out, I noticed that the version is retained from the active resource. I think we will want to modify the draft resource to remove the version, right? I saw Matt's comment above about creating a new task for limiting the amount of drafts that get created. Not sure if we want to include the version removal in that task or in this one

app/src/pages/authoring/index.tsx
```diff
@@ -65,6 +97,25 @@ export default function AuthoringPage() {
           <Button w={240} loading={draftMutation.isLoading} onClick={createResource}>
             {`Create New Draft ${resourceType}`}
           </Button>
+          <Title order={3}>{`Start From an Existing ${resourceType}:`}</Title>
+          {artifacts ? (
+            <Select
+              label={`Select an existing ${resourceType} to create a draft from`}
```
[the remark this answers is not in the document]
elsaperelli: Oh gosh grammar 😅

app/src/server/trpc/routers/service.ts
elsaperelli: Ahh for some reason I thought this was not doing that but I think you are right! Not quite sure how to work around this, I will look into it

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -29,14 +36,49 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+
+  const createDraftOfArtifact = () => {
+    const draftOfArtifact = jsonData;
+    draftOfArtifact.id = uuidv4();
+    draftOfArtifact.status = 'draft';
+    draftMutation.mutate({ resourceType, draft: draftOfArtifact });
+  };
```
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/[resourceType]/[id].tsx]: Hmm, good suggestion that I didn't think about. I am leaning toward keeping `createDraft` as is. I think it's best if `createDraft` is kept as simple as possible so that we can reuse it in multiple places

app/src/pages/authoring/index.tsx
```diff
@@ -16,39 +27,74 @@ const useStyles = createStyles(() => ({
...
+    }
+  });
 
   const createResource = () => {
     const newResource = resourceType === 'Measure' ? { ...MeasureSkeleton } : { ...LibrarySkeleton };
     newResource.id = uuidv4();
     draftMutation.mutate({ resourceType, draft: newResource });
   };
 
+  const createDraftArtifactOf = () => {
```
sarahmcdougall, line 92: This is a nitpick but since this `const` does not take in any arguments, I feel like it might be more appropriate to name this `createDraftArtifact` rather than saying "of" at the end. Or we could make it more descriptive and call it `createDraftArtifactOfResource`

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/[resourceType]/[id].tsx]: I did notice that we have the same code in `app/src/service/trpc/routers/service` and `app/src/pages/[resourceType]/[id].tsx` for setting the id, status, and deleting the version, so maybe a compromise here would be to create a helper function under the `util` directory called `modifyResourceFields` (or something similar) that takes care of these modifications. But we aren't duplicating that much code so it's not a big deal

sarahmcdougall (commented):
Just a few comments that I wanted to bring up - otherwise looks good! The changes look great

app/src/pages/authoring/index.tsx
```diff
@@ -65,6 +111,31 @@ export default function AuthoringPage() {
...
+            />
+          ) : (
+            <Text c="red">An unknown error occurred fetching artifacts</Text>
+          )}
+          <Button
+            w={240}
+            loading={draftFromArtifactMutation.isLoading}
+            onClick={createDraftArtifactOf}
+            disabled={!selectedArtifact}
+          >
```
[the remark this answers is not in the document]
elsaperelli: Oo good point! I think I'll just change this so that the button is always rendered and disabled unless a resource is selected.

BriannaVH (commented):
Just one comment about something that is pretty subjective. Other than that, everything looks great to me!

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -1,18 +1,25 @@
...
-import { Divider, Group, Space, Stack, Tabs, Text } from '@mantine/core';
+import { Button, Divider, Group, Space, Stack, Tabs, Text } from '@mantine/core';
 import { GetServerSideProps, InferGetServerSidePropsType } from 'next';
 import { FhirArtifact } from '@/util/types/fhir';
 import { useEffect, useMemo } from 'react';
 import CQLRegex from '../../util/prismCQL';
 import { Prism as PrismRenderer } from 'prism-react-renderer';
 import parse from 'html-react-parser';
+import { v4 as uuidv4 } from 'uuid';
+import { AlertCircle, CircleCheck } from 'tabler-icons-react';
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx:9]: LOL I love it. I think I'll keep the AlertCircle for now since that was the precedent that was set- but maybe that'll change in the future :)

sarahmcdougall (approved):
Great work!

BriannaVH (approved):
Everything looks great to me!

PR #41, opened 2023-06-12
Render data requirements

BriannaVH:
# Summary
Renders the data requirements tab and populates the panel of the selected resource when a button is clicked

## New behavior
Users can now click a button and render a data requirements tab with a specified resource's respective data requirements. If the fetch was successful, a drop down notification will appear to inform the user of the success and the tab/panel will render on the page. Otherwise, if no data requirements exist for the resource, a drop down notification will appear informing the user of this and the tab should not appear. 

## Code changes

- I imported several new mantine features such as:  `IconAbacusOff,` `IconCheck,` and `notifications` so I could utilize these features in the UI.
- I also imported `useState` from react so I could update different features in real time. 
- Added a mantine Button which, when clicked, fetches the data requirements, displays a notification on whether the fetch was successful or not, and renders the tab/panel on the page.

# Testing guidance
```
npm run check:all
 npm run start:all 
```
Navigate to `http://localhost:3001/` and select both Measure and Library packages. For any measure/library resource, click the `Get Data Requirements button`. For resources without data requirements (FHIR Library for instance), a notification should appear informing the user that there are no data requirements and the new tab should NOT render. Now, select a library with data requirements such as HospiceFHIR4 and click the `get Data requirements button`. The tab should render with the populated information and another drop down notification should inform the user of the successful fetch.

### DIFF
--- a/app/src/pages/[resourceType]/[id].tsx
+++ b/app/src/pages/[resourceType]/[id].tsx
@@ -1,11 +1,17 @@
 import { Prism } from '@mantine/prism';
-import { Divider, Group, Space, Stack, Tabs, Text } from '@mantine/core';
+import { Divider, Group, Space, Stack, Tabs, Text, Button } from '@mantine/core';
+import { IconAbacus, IconAbacusOff, IconCheck } from '@tabler/icons-react';
+import { notifications } from '@mantine/notifications';
+import React from 'react';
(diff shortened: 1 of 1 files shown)


elsaperelli:
If you want to add the notification icons back, I think you just need to import them from `tabler-icons-react`

PR #42, opened 2023-06-13
Render draft artifacts in authoring mode

sarahmcdougall:
# Summary
Builds out the `/authoring/[resourceType].tsx` page, allowing the user to view draft artifacts as cards that route to the editing page.

## New behavior
The user can now view resources cards of the available draft artifacts that are stored in the mongo database, which contain identifying fields like `url` and `identifier`. To do so, from authoring “mode,” select a resource type. Then, the contents of each draft resource should be rendered as a resource card. Each resource card also contains an edit button that routes to the editing page so that the user can edit the selected resource.
Screenshot 2023-06-12 at 3 00 11 PM

## Code changes
* `app/src/pages/authoring/[resourceType]/index.tsx` —> builds out the `[resourceType]` page by getting the resources using a `trpc` query, mapping the resources to the resource card content (`id`, `identifier`, etc.), and rendering the resource cards on the page
* `app/src/components/ResourceCards.tsx` —> changed styling on the `Stack` so that the cards have uniform width with different screen sizes (previously this wasn’t the case)
* Changes to the resource card props
    * Added an `icon` field for the icon to render on each card — should be an external link for the repository “mode” and a pencil/edit icon for the authoring “mode”
    * Added an `authoring` boolean field that gets checked to determine what page to route to (depending on the mode) and what tooltip text to render
* Added tooltip for editing a draft resource/viewing resource contents (depending on the mode)
* `app/src/util/resourceCardUtils.ts` —> new file with a function for extracting the fields that we want to render on a resource card. I noticed that we do this extraction in multiple places, so moving this into a helper seemed like a good improvement

# Testing guidance
Test the various scenarios:
* Clear the `Measure` and `Library` collections on the `draft-repository` database so that each resource type button is at “0.” There should be no resource cards on the page - you should see “No <resourceType> resources available.”
Screenshot 2023-06-12 at 2 57 47 PM

* Create a few draft resources for each resource type. Then, click on the resource type button. You should see a resource card of reach draft resource.
* Try creating a few resources that have a `url` and/or `identifier` field - these fields will be reflected on the resource cards when they are defined.
* Hover over the “edit” button - you should see the text “Edit Draft Resource”
* Click on the “edit” button and check that you get routed to the editing page, and that the resource corresponds to the resource card that was clicked on. Try editing a field and check that the change is reflected on the card
* Change the window size and check the responsiveness of the resource cards - they should now be of the same width
Screenshot 2023-06-12 at 3 00 31 PM

### DIFF
--- a/app/src/components/ResourceCards.tsx
+++ b/app/src/components/ResourceCards.tsx
@@ -29,7 +29,7 @@ export default function ResourceCards({ resourceInfo, resourceType, authoring }:
       <ScrollArea.Autosize mah={height * 0.8} type="scroll">
         {resourceInfo?.length > 0 ? (
           <Center>
-            <Stack align="stretch">
+            <Stack align="stretch" pb={2}>
               {resourceInfo.map(res => {
                 return (
                   <div key={res.id}>
--- a/app/src/components/ResourceCards.tsx
+++ b/app/src/components/ResourceCards.tsx
@@ -3,16 +3,17 @@ import { Center, ScrollArea, Stack, Text } from '@mantine/core';
 import { useEffect, useState } from 'react';
 import ResourceInfoCard from './ResourceInfoCard';
 
-interface ResourceButtonsProps {
+interface ResourceCardsProps {
   resourceInfo: ResourceInfo[];
   resourceType: ArtifactResourceType;
+  icon: JSX.Element;
 }
 
 /**
  * Component which displays all resources of a specified type and their ids
  * as buttons that link to that resource's page
  */
-export default function ResourceCards({ resourceInfo, resourceType }: ResourceButtonsProps) {
+export default function ResourceCards({ resourceInfo, resourceType, icon }: ResourceCardsProps) {
   const [height, setWindowHeight] = useState(0);
   useEffect(() => {
     const handleResize = () => {
@@ -26,7 +27,7 @@ export default function ResourceCards({ resourceInfo, resourceType }: ResourceBu
   return (
     <div>
       <ScrollArea.Autosize mah={height * 0.8} type="scroll">
-        {resourceInfo.length > 0 ? (
+        {resourceInfo?.length > 0 ? (
           <Stack
             style={{
               alignItems: 'center',
@@ -36,7 +37,7 @@ export default function ResourceCards({ resourceInfo, resourceType }: ResourceBu
             {resourceInfo.map(res => {
               return (
                 <div key={res.id}>
-                  <ResourceInfoCard resourceInfo={res} />
+                  <ResourceInfoCard resourceInfo={res} icon={icon} />
                 </div>
               );
             })}
--- a/app/src/components/ResourceCards.tsx
+++ b/app/src/components/ResourceCards.tsx
@@ -12,7 +12,7 @@ interface ResourceCardsProps {
 
 /**
  * Component which displays all resources of a specified type and their ids
- * as buttons that link to that resource's page
+ * as resource cards that link to that resource's page
  */
 export default function ResourceCards({ resourceInfo, resourceType, icon, authoring }: ResourceCardsProps) {
   const [height, setWindowHeight] = useState(0);
@@ -29,15 +29,17 @@ export default function ResourceCards({ resourceInfo, resourceType, icon, author
     <div>
       <ScrollArea.Autosize mah={height * 0.8} type="scroll">
         {resourceInfo?.length > 0 ? (
-          <Stack align="stretch">
-            {resourceInfo.map(res => {
-              return (
-                <div key={res.id}>
-                  <ResourceInfoCard resourceInfo={res} icon={icon} authoring={authoring} />
-                </div>
-              );
-            })}
-          </Stack>
+          <Center>
+            <Stack align="stretch">
+              {resourceInfo.map(res => {
+                return (
+                  <div key={res.id}>
+                    <ResourceInfoCard resourceInfo={res} icon={icon} authoring={authoring} />
+                  </div>
+                );
+              })}
+            </Stack>
+          </Center>
         ) : (
           <Center>
             <Text>
--- a/app/src/components/ResourceCards.tsx
+++ b/app/src/components/ResourceCards.tsx
@@ -6,7 +6,6 @@ import ResourceInfoCard from './ResourceInfoCard';
 interface ResourceCardsProps {
   resourceInfo: ResourceInfo[];
   resourceType: ArtifactResourceType;
-  icon: JSX.Element;
   authoring?: boolean;
 }
 
@@ -14,7 +13,7 @@ interface ResourceCardsProps {
  * Component which displays all resources of a specified type and their ids
  * as resource cards that link to that resource's page
  */
-export default function ResourceCards({ resourceInfo, resourceType, icon, authoring }: ResourceCardsProps) {
+export default function ResourceCards({ resourceInfo, resourceType, authoring }: ResourceCardsProps) {
   const [height, setWindowHeight] = useState(0);
   useEffect(() => {
     const handleResize = () => {
@@ -34,7 +33,7 @@ export default function ResourceCards({ resourceInfo, resourceType, icon, author
               {resourceInfo.map(res => {
                 return (
                   <div key={res.id}>
-                    <ResourceInfoCard resourceInfo={res} icon={icon} authoring={authoring} />
+                    <ResourceInfoCard resourceInfo={res} authoring={authoring} />
                   </div>
                 );
               })}
--- a/app/src/components/ResourceCards.tsx
+++ b/app/src/components/ResourceCards.tsx
@@ -29,12 +29,7 @@ export default function ResourceCards({ resourceInfo, resourceType, icon, author
     <div>
       <ScrollArea.Autosize mah={height * 0.8} type="scroll">
         {resourceInfo?.length > 0 ? (
-          <Stack
-            style={{
-              alignItems: 'center',
-              paddingBottom: '12px'
-            }}
-          >
+          <Stack align="stretch">
             {resourceInfo.map(res => {
               return (
                 <div key={res.id}>
--- a/app/src/components/ResourceCards.tsx
+++ b/app/src/components/ResourceCards.tsx
@@ -7,13 +7,14 @@ interface ResourceCardsProps {
   resourceInfo: ResourceInfo[];
   resourceType: ArtifactResourceType;
   icon: JSX.Element;
+  authoring?: boolean;
 }
 
 /**
  * Component which displays all resources of a specified type and their ids
  * as buttons that link to that resource's page
  */
-export default function ResourceCards({ resourceInfo, resourceType, icon }: ResourceCardsProps) {
+export default function ResourceCards({ resourceInfo, resourceType, icon, authoring }: ResourceCardsProps) {
   const [height, setWindowHeight] = useState(0);
   useEffect(() => {
     const handleResize = () => {
@@ -37,7 +38,7 @@ export default function ResourceCards({ resourceInfo, resourceType, icon }: Reso
             {resourceInfo.map(res => {
               return (
                 <div key={res.id}>
-                  <ResourceInfoCard resourceInfo={res} icon={icon} />
+                  <ResourceInfoCard resourceInfo={res} icon={icon} authoring={authoring} />
                 </div>
               );
             })}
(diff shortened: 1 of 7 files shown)


app/src/util/resourceCardUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { FhirArtifact, ResourceInfo } from './types/fhir';
+
+/**
+ * Extracts fields from the resource to be rendered onto the resource's card.
+ * @param resource Measure of Library resource
```
elsaperelli, line 5:
```suggestion
 * @param resource Measure or Library resource
```

app/src/components/ResourceInfoCard.tsx
```diff
@@ -1,11 +1,12 @@
-import { Button, Grid, Paper, Text, createStyles, em, getBreakpointValue, rem } from '@mantine/core';
+import { ActionIcon, Grid, Paper, Text, createStyles, em, getBreakpointValue, rem, Tooltip } from '@mantine/core';
 import Link from 'next/link';
 import React from 'react';
 import { ResourceInfo } from '@/util/types/fhir';
-import { ExternalLink } from 'tabler-icons-react';
 
 export interface ResourceInfoCardProps {
   resourceInfo: ResourceInfo;
+  icon: JSX.Element;
```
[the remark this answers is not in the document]
sarahmcdougall: I like this idea :) I had added the `icon` property before realizing I also needed to have dynamic routing based on the mode - being able to use the `authoring` property for determining both the icon and the route makes the logic easier to understand 👍

app/src/pages/[resourceType]/search-result.tsx
```diff
@@ -24,9 +26,11 @@ export default function ResourceSearchResultsPage({
...
       <Divider my="md" />
       {resourceInfo ? (
-        <div style={{ paddingTop: '18px' }}>
-          <ResourceCards resourceInfo={resourceInfo} resourceType={resourceType} />
-        </div>
+        <Center>
+          <div style={{ paddingTop: '18px' }}>
+            <ResourceCards resourceInfo={resourceInfo} resourceType={resourceType} icon={<ExternalLink size="24" />} />
+          </div>
+        </Center>
```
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/[resourceType]/search-result.tsx]: Good catch! Look at us maximizing the benefits of using Mantine 😍

app/src/pages/authoring/[resourceType]/index.tsx
```diff
@@ -1,10 +1,35 @@
...
+import { useRouter } from 'next/router';
+import { ArtifactResourceType, FhirArtifact, ResourceInfo } from '@/util/types/fhir';
+import ResourceCards from '@/components/ResourceCards';
+import { extractResourceInfo } from '@/util/resourceCardUtils';
+
 export default function ResourceAuthoringPage() {
+  const router = useRouter();
+  const { resourceType } = router.query;
+
+  const resourceTypeQuery = trpc.draft.getDrafts.useQuery(resourceType as ArtifactResourceType);
```
Comment (mgramigna) [inline app/src/pages/authoring/[resourceType]/index.tsx:12]: Same issue as https://github.com/projecttacoma/measure-repository/pull/40#discussion_r1230012363


image

app/src/pages/authoring/[resourceType]/index.tsx
```diff
@@ -1,10 +1,35 @@
...
+
+  const resourceCardContent: ResourceInfo[] = (resourceTypeQuery.data as FhirArtifact[])?.map(resource => {
+    return extractResourceInfo(resource);
+  });
```
Comment (mgramigna) [inline app/src/pages/authoring/[resourceType]/index.tsx:16]: I think it makes more sense to do this:

```suggestion
  const resourceCardContent: ResourceInfo[] = (resourceTypeQuery.data ?? []).map(resource => {
    return extractResourceInfo(resource);
  });
```

And have it be empty array when data is undefined.

Also, we should probably memo-ize this, something like this:

```ts
  const {data: artifacts} = trpc.draft.getDrafts.useQuery(resourceType as ArtifactResourceType);

  const resourceCardContent = useMemo(() => {
    return (artifacts ?? []).map(a => extractResourceInfo(a));
  }, [artifacts])

```

mgramigna (changes requested):

app/src/components/ResourceCards.tsx
```diff
@@ -26,21 +27,18 @@ export default function ResourceCards({ resourceInfo, resourceType }: ResourceBu
...
-              return (
-                <div key={res.id}>
-                  <ResourceInfoCard resourceInfo={res} />
-                </div>
-              );
-            })}
-          </Stack>
+        {resourceInfo?.length > 0 ? (
+          <Center>
+            <Stack align="stretch" pb={2}>
```
elsaperelli, line 32: I think we can do what you did in `[resourceType].tsx` and change this to just be `Slack align="center"`, unless you want stretch but I am not entirely sure of its purpose. Also, I am not sure if `pb={2}` is necessary? I couldn't see any visual difference without it but maybe I am wrong!

elsaperelli (changes requested):
Nice! Just one more small comment. Also, I still notice the scrolling with the resource cards is different between the "Repository" and "Authoring" mode and I think it has to do with how the ScrollArea height is set in `ResourceCards.tsx` since in "Repository" mode has a search button...are we going to have a search button for the draft artifacts? If so, then we can leave this and when it's added it'll be the same, but if not, maybe add a parameter to ResourceCards that indicates if we want a search button or not and then calculate the scroll height based on that? Does that make sense? This can also be more of an EOS activity since it is totally related to your PR.

app/src/components/ResourceCards.tsx
sarahmcdougall: The visual difference with the bottom padding is very slight. It makes sure that the last card in the `Stack` does not get cut off (which essentially means that its shadow will still appear on the page).

Without the bottom padding (look at the card for `Measure/123`):
Screenshot 2023-06-16 at 1 38 42 PM

With the bottom padding:
Screenshot 2023-06-16 at 1 39 18 PM

With `align="center"`, this is what happens sometimes when working with a condensed screen size:
Screenshot 2023-06-16 at 2 39 43 PM

For the most part we won't see this I guess, but I found that using `align="stretch"` stretches the resource cards to take up the entire width so that they are uniform width no matter the screen size:
Screenshot 2023-06-16 at 2 41 23 PM

sarahmcdougall:
> Nice! Just one more small comment. Also, I still notice the scrolling with the resource cards is different between the "Repository" and "Authoring" mode and I think it has to do with how the ScrollArea height is set in `ResourceCards.tsx` since in "Repository" mode has a search button...are we going to have a search button for the draft artifacts? If so, then we can leave this and when it's added it'll be the same, but if not, maybe add a parameter to ResourceCards that indicates if we want a search button or not and then calculate the scroll height based on that? Does that make sense? This can also be more of an EOS activity since it is totally related to your PR.

Yeah the scrolling is different right now because of the search button - I think that we might want to be able to search for the draft artifacts in the future as a nice-to-have feature

app/src/components/ResourceInfoCard.tsx
```diff
@@ -1,11 +1,12 @@
-import { Button, Grid, Paper, Text, createStyles, em, getBreakpointValue, rem } from '@mantine/core';
+import { ActionIcon, Grid, Paper, Text, createStyles, em, getBreakpointValue, rem, Tooltip } from '@mantine/core';
 import Link from 'next/link';
 import React from 'react';
 import { ResourceInfo } from '@/util/types/fhir';
-import { ExternalLink } from 'tabler-icons-react';
+import { Edit, ExternalLink } from 'tabler-icons-react';
```
lmd59, line 5: I know ExternalLink was a previous PR icon choice, but doesn't it imply that... clicking it will link externally? Maybe we could use something like Details instead? https://tabler-icons.io/i/details

sarahmcdougall: I agree - it makes sense to update this since the user isn't being taken to an external link.

I tried out a few different icons - let me know your thoughts!

`Details`:
Screenshot 2023-06-19 at 9 47 09 AM

`DotsCircleHorizontal`:
Screenshot 2023-06-19 at 9 50 13 AM

`ChevronRight`:
Screenshot 2023-06-19 at 9 52 22 AM

`SquareArrowRight`:
Screenshot 2023-06-19 at 9 56 43 AM

`ArrowBarRight`:
Screenshot 2023-06-19 at 9 57 36 AM

I am leaning toward using an icon that includes a right arrow since we are routing to a new page, though it's not an external link. I'm not totally against using the `Details` icon but it is not intuitive to me just from looking at it that it routes to a new page

app/src/pages/authoring/[resourceType]/index.tsx
```diff
@@ -1,10 +1,36 @@
...
+
   return (
     <div>
       <Center>
-        <Text c="gray">Coming Soon</Text>
+        <Text c="gray" fz="xl">
+          Available Draft {resourceType} Resources
+        </Text>
       </Center>
+      <Divider my="md" />
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/index.tsx:26]: Not sure if this is a matter of importance or not, but I noticed on both my smaller/bigger monitor that the` "Available Draft {resourceType} Resouces" header` along with the cards seem to be slightly off-centered from the `"Repository/Authoring" tabs` in the navigation bar at the top. This might not matter if we are planning on adding more tabs to the navigation bar, but maybe we could move the header and the cards so that they are centered underneath those two tabs? 

Here is what the image looks like on my screen:
UncenteredHeader

app/src/components/ResourceInfoCard.tsx
BriannaVH: I also like `SquareArrowRight` along with `DotsCircleHorizontal`

app/src/pages/[resourceType]/search-result.tsx
```diff
@@ -24,9 +25,9 @@ export default function ResourceSearchResultsPage({
       </Center>
```
Comment (BriannaVH) [inline app/src/pages/[resourceType]/search-result.tsx:25]: The `"Search Returned" header` is also a bit off-centered from the two tabs (Repository/Authoring) in the navigation bar.

app/src/pages/[resourceType]/search-result.tsx
```diff
@@ -75,16 +76,7 @@ export const getServerSideProps: GetServerSideProps<{
...
-        const resourceInfo: ResourceInfo = {
-          resourceType: checkedResourceType,
-          id: entry.resource.id,
-          identifier: identifier?.system && identifier?.value ? `${identifier.system}|${identifier.value}` : null,
-          name: entry.resource.name ?? null,
-          url: entry.resource.url ?? null,
-          version: entry.resource.version ?? null,
-          status: entry.resource.status ?? null
-        };
+        const resourceInfo = extractResourceInfo(entry.resource);
```
Comment (BriannaVH) [inline app/src/pages/[resourceType]/search-result.tsx:79]: Currently, should the search functionality also take into consideration draft artifacts?

Comment (sarahmcdougall) [reply app/src/pages/[resourceType]/search-result.tsx:79]: Right now, the search functionality should only take artifacts from the service into account. In the future, we might want to add a separate search for drafts, or even combine search results for drafts and active artifacts, but I think that's outside of the scope of this task because it will require some more discussion to figure out exactly what we want to do. We have a task in the backlog for refactoring some of the search results to de-duplicate some of the code, so we could bring this up when we start exploring that task.

app/src/pages/authoring/[resourceType]/index.tsx
Comment (sarahmcdougall) [reply app/src/pages/authoring/[resourceType]/index.tsx:26]: I think the reason why we are seeing this is because the "Repository"/"Authoring" tabs are centered in respect to the entire screen, whereas the "Available Draft Library Resources" text is centered in respect to the `index.tsx` page, which is separate from the sidebar that we see on the left. If we were to change the centering here, we would need to do the same on a few other pages (like the search results, which you mentioned). Since the precedent was set to have the text centered in respect to the page rather than the entire screen, I think I will keep this as-is for now and we can consider making a task to update the styling across all the necessary pages

Comment (lmd59) [reply app/src/pages/authoring/[resourceType]/index.tsx:26]: This one's tricky because the tabs are centered within the top bar, whereas the header is centered within its resource list section (which is offset from the top bar by the navigation on the left). Because the tabs also change the navigation, they should remain within the top bar. I don't mind the different centeredness too much, but if others find it jarring, the tabs could potentially be moved to be left aligned so that the different centeredness isn't a bother. Next to the main Measure Repository label/link, something like this:
Screen Shot 2023-06-20 at 12 03 14 PM

Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/index.tsx:26]: Honestly, I prefer it centered like it was, but I don't have great reason for that other than it's my gut feel :)

elsaperelli (approved):
👏

app/src/pages/authoring/[resourceType]/index.tsx
Comment (BriannaVH) [reply app/src/pages/authoring/[resourceType]/index.tsx:26]: Sounds good to me! This is really more of a subjective aspect of the UI so it's more up to personal preference than anything :)

BriannaVH (approved):
Looks great!

PR #43, opened 2023-06-16
Render data requirements for a selected resource

BriannaVH:
# Summary
This PR allows users to selectively fetch and render the data requirements of a specific resource by clicking on a button.

## New behavior
Users can now click the `get data requirements button` to fetch the data requirements of a specific resource(Measure or Library). If the fetch was successful and the resource has data requirements to display, a successful pop up notification will appear and the `data requirements tab/panel `will render populated with its respective content. When the user navigates away from the page after a successful fetch and returns, the data requirements tab and content should still be rendered. If the resource doesn't have data requirements at all, a pop up notification will appear informing the user of this and the tab/panel should not render. 

## Code changes

- Created a new trpc procedure in the `app/src/server/trpc/routers/service.ts` file called `getDataRequirements` that fetches the data requirements of a specified resource.
- Called a useQuery function in `app/src/pages/[resourceType]/[id].tsx`  that starts off as disabled so it doesn't run until triggered.
- Created a mantine button in `app/src/pages/[resourceType]/[id].tsx` that when clicked, refetches the above useQuery hook so that it runs.
- Created a `dataRequirements` variable that becomes populated with the resource's data requirements and determines whether the tab/panel should render on the page.

# Testing guidance
```
npm run check:all 
npm run start:all
```
 Navigate to http://localhost:3001/ and select a combination of Measure and Library resources. For each resource, click the “Get Data Requirements” button. If there are no data requirements (`FHIR Library` is an example), a notification should appear notifying the user that no data requirements are present for this resource and the `data requirements tab/panel` should NOT appear. Now, select a library with data requirements such as` HospiceFHIR4` and click the `get Data requirements button`. The button should appear as if it is loading, then return to its normal state, and a data requirements tab should appear with the populated information. After a successful fetch, leave the page and then return to the HospiceFHIR4 library, the data requirements tab/panel should still be rendered and populated. No matter how many times the button is clicked on after this, regardless of whether the package has data requirements or not, the exact same functionality should repeat.


app/src/pages/[resourceType]/[id].tsx
```diff
@@ -28,6 +31,45 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     /* eslint-enable @typescript-eslint/no-explicit-any */
   }, []);
 
+  const [loadingIconVisible, setLoadingIconVisible] = useState(false);
+  const { data: dataRequirements, refetch } = trpc.service.getDataRequirements.useQuery(
+    { resourceType: jsonData.resourceType, id: jsonData.id as string },
+    { enabled: false }
+  );
```
Comment (sarahmcdougall) [inline app/src/pages/[resourceType]/[id].tsx:38]: Small suggestion here - I've noticed that it is common practice to put `const`s at the beginning of the function, above the `useEffect`(s).  (See [app/src/pages/authoring/[resourceType]/[id].tsx](https://github.com/projecttacoma/measure-repository/blob/main/app/src/pages/authoring/%5BresourceType%5D/%5Bid%5D.tsx) as an example). I think we should put these two `const`s above the `useEffect` for the custom Prism rendering

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -28,6 +31,45 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+  //useEffect checks if the data Requirements were loaded and whether the button was clicked
+  useEffect(() => {
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx:40]: Actually, we probably want to get rid of the `useEffect()` anyways. You can actually utilize the `onSuccess` and `onError` properties on `useQuery()` to show the notifications! Check out how we did this [here](https://github.com/projecttacoma/measure-repository/blob/main/app/src/pages/%5BresourceType%5D/%5Bid%5D.tsx#L42). You can add it right where you set `enabled: false` so it'll look like: 

```
{ enabled: false, onSuccess: () => {notifications.show({....}), onError: () => {notifications.show({...})
```

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -79,41 +121,66 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+                </Prism>
+              </Tabs.Panel>
+            )}
+            {jsonData.text && (
+              <Tabs.Panel value="narrative">
+                <Space h="sm" />
+                {parse(jsonData.text.div)}
+              </Tabs.Panel>
+            )}
+            {dataRequirements?.Library.resourceType == 'Library' && (
```
Comment (elsaperelli) [inline app/src/pages/[resourceType]/[id].tsx:162]: If you make the change below, this will change to just be `dataRequirements?.resourceType === 'Library'` I believe. Also, it's good practice to use strict equality (`===`). You can read more about that [here](https://www.tektutorialshub.com/typescript/strict-equality-loose-equality-in-typescript/).

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -79,41 +121,66 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+              <Tabs.Panel value="datarequirements">
```
Comment (elsaperelli) [inline app/src/pages/[resourceType]/[id].tsx:163]:
```suggestion
              <Tabs.Panel value="data-requirements">
```

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -79,41 +121,66 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+                <Prism language="json" colorScheme="light">
+                  {JSON.stringify(dataRequirements, null, 2)}
+                </Prism>
+              </Tabs.Panel>
+            )}
+          </Tabs>
+        </div>
+        <div style={{ position: 'absolute', left: '85vw', top: '16vh' }}>
```
Comment (elsaperelli) [inline app/src/pages/[resourceType]/[id].tsx:171]: I would maybe change the placement of the button to be next to the `Create Draft of` button.

elsaperelli (changes requested):
Great work with this!!! Just some comments about best practices and some other things I recently discovered we can do differently by utilizing `useQuery` :)

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -28,6 +31,45 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+            id: 'no-requirements',
+            withCloseButton: true,
+            autoClose: 3000,
+            title: 'No Data Requirements Found',
+            message: 'No data requirements were found for this resource',
+            color: 'red',
+            style: { backgroundColor: 'white' },
+            icon: <AbacusOff />,
+            loading: false
+          });
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx:66]: I also think we may have to change some things around with error handling if we want to utilize `onSuccess`, `onError`, and `isLoading`. I think we should utilize trpc error handling within the `getDataRequirements` procedure so that we can actually use these properties. Right now, we cannot use them because the fetch is always successful even if it is an Operation outcome, it is still returning something. Maybe we can make these changes in the procedure in `service.ts` by utilizing [trpc errors](https://trpc.io/docs/server/error-handling). Basically we can check what the details.text is and throw a TRPC error based on what kind of error it is. I haven't actually tried this, but I think it'll work.

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -79,41 +121,66 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+            id="btn"
+            loading={loadingIconVisible}
+            loaderPosition="center"
+            onClick={() => {
+              refetch();
+              setLoadingIconVisible(true);
+            }}
+          >
+            Get Data Requirements
+          </Button>
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: Thats a great idea and a good jumping off point! I think this may be because I am on my smaller monitor, but I think it actually looks better for the buttons to be side-by-side than stacked on top of one another. However I don't have my bigger monitor on me right now, so please let me know if the side-by-side doesn't look that great and should be modified for a varying screen size.

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx]: Sorry I should've been a little bit clearer on this! Some other things have to chance in order for `onSuccess` and `onError` to work. See my comment [here](https://github.com/projecttacoma/measure-repository/pull/43#discussion_r1234392601) about that!

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -28,6 +31,45 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     /* eslint-enable @typescript-eslint/no-explicit-any */
   }, []);
 
+  const [loadingIconVisible, setLoadingIconVisible] = useState(false);
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx]: I tested this a bit, and we actually want to use the `isFetching` property on `useQuery` rather than `isLoading`. This is because we set enabled to false, so what we actually want is for the loading to occur when the fetching is happening, not when dataRequirements is not populated.

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/[resourceType]/[id].tsx]: They look good side-by-side on my laptop and on my monitor 👍 looks better than having them stacked in my opinion

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: That was a great suggestion, thank you! I implemented this and it worked :)

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: Just implemented your second suggestion Sarah!So now if that same error were to pop up it should now show you the `"text"` portion of the resource. Thank you both for your great suggestions!

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -73,47 +110,72 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
             <Text size="xl" color="gray">
               {jsonData.resourceType}/{jsonData.id}
             </Text>
-            <Button w={240} loading={draftMutation.isLoading} onClick={createDraftOfArtifact}>
-              Create Draft of {jsonData.resourceType}
-            </Button>
+            <Group>
+              <Button
+                w={240}
+                id="btn"
```
Comment (sarahmcdougall) [inline app/src/pages/[resourceType]/[id].tsx:116]:
```suggestion
```
I think we can safely get rid of this since we aren't doing any sort of unit testing for this app where an id would need to be specified

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -73,47 +110,72 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
-              {/* eslint-disable  @typescript-eslint/no-explicit-any */}
-              <Prism language={'cql' as any} colorScheme="light">
-                {/* eslint-enable  @typescript-eslint/no-explicit-any */}
-                {decodedCql}
-              </Prism>
-            </Tabs.Panel>
-          )}
-          {decodedElm != null && (
-            <Tabs.Panel value="elm" pt="xs">
+        <div>
```
Comment (sarahmcdougall) [inline app/src/pages/[resourceType]/[id].tsx:132]: Now that I'm taking a closer look at the code diff in this file, I notice that it looks like a lot has changed with the tab code when really it's just that we moved all the tab code inside of a `<div>`. I don't think we need to wrap this all in a div (I didn't get any errors and none of the functionality changed) so I think we can remove the `<div>` tag on this line and on line 178

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/[resourceType]/[id].tsx]: Screenshot 2023-06-20 at 12 32 16 PM
Woohoo! This is a great change!

sarahmcdougall (changes requested):

elsaperelli:
This is probably a question more for @mgramigna or @sarahmcdougall, but I noticed that the data requirements output for Library/FHIRHelpers is a Library (not an Operation Outcome) but that the data-requirements property on this Library is empty. Is this okay to display? Or should we have an additional check that the data-requirements property is defined on the Library?

app/src/server/trpc/routers/service.ts
```diff
@@ -36,6 +37,25 @@ export const serviceRouter = router({
...
+    .input(z.object({ resourceType: z.enum(['Measure', 'Library']), id: z.string() }))
+    .query(async ({ input }) => {
+      const res = await fetch(
+        `${process.env.NEXT_PUBLIC_MRS_SERVER}/${input.resourceType}/${input.id}/$data-requirements`
+      );
+      const resource = await res.json();
+      if (resource?.resourceType === 'OperationOutcome') {
+        const eMessage = resource?.issue[0]?.details?.text;
+        throw new TRPCError({
+          code: 'INTERNAL_SERVER_ERROR',
```
elsaperelli, line 50: So you can also make the `code` here flexible since it is something that would be on the operation outcome. Try accessing it like you access the message on `resource?.issue[0]?.details?.text`!

app/src/server/trpc/routers/service.ts
```diff
@@ -36,6 +37,25 @@ export const serviceRouter = router({
...
+          message: eMessage
+        });
+      }
+      return {
+        Library: resource as fhir4.Library
+      } as const;
```
elsaperelli, line 56: In other procedures, it was helpful to return an object since there were multiple things, but here since we are just returning a Library I think you can just do `return resource as fhir4.Library`. You will need to change the way you access `resourceType` in the other file.

app/src/server/trpc/routers/service.ts
```diff
@@ -14,7 +16,6 @@ export const serviceRouter = router({
     ]).then(([resMeasure, resLibrary]) =>
       Promise.all([resMeasure.json() as Promise<fhir4.Bundle>, resLibrary.json() as Promise<fhir4.Bundle>])
     );
```
elsaperelli, line 18:
```suggestion
    );
    
```

elsaperelli (changes requested):

sarahmcdougall:
> This is probably a question more for @mgramigna or @sarahmcdougall, but I noticed that the data requirements output for Library/FHIRHelpers is a Library (not an Operation Outcome) but that the data-requirements property on this Library is empty. Is this okay to display? Or should we have an additional check that the data-requirements property is defined on the Library?

I think Matt and I briefly talked about this at some point but didn't come to a conclusion on how to handle this. I think that for now it might be fine to just keep things as is, and then when we make things more human-readable (in future tasking) we can try to tackle this condition. Right now I think it's fine to just display whatever comes back from the operation, whether it has a populated `dataRequirement` array or not

app/src/server/trpc/routers/service.ts
BriannaVH: I'm a bit confused, do you want me to directly set the message variable to resource?.issue[0]?.details?.text or something else?

elsaperelli: I'm suggesting that rather than hardcoding code to be 'INTERNAL_SERVER_ERROR', you can have it be equal to whatever `code` is on the OperationOutcome. So, `code: resource?.issue[0]?.code`

BriannaVH: ohhh, okay thank you!

elsaperelli (approved):
Great job with this task! Looking forward to seeing more iterations of this! Thanks for being patient and flexible with change requests/suggestions :)

sarahmcdougall (approved):
Great work on this!!

PR #44, opened 2023-06-20
Edit draft artifacts

elsaperelli:
# Summary
This PR allows the user to be able to edit/modify existing draft artifacts and save those edits to the local database.

## New behavior
When a user creates or selects a draft artifact, a list of text inputs for the following fields are displayed next to the JSON of the artifact: `url`, `identifier`, `name`, `title`, and `description`. If any of these fields already exists on the artifact, then the text inputs will be populated with their current values. The user can edit or delete the content of these fields, but artifact in the local database will only be updated when the user clicks "Submit". 

## Code changes
- `app/src/pages/authoring/[resourceType]/[id].tsx` - Add inputs for `name`, `title`, and `description`, modified the `parseUpdate` function to return additions and deletions that we want to update the artifact with, rather than just additions, add an 'X' ActionIcon to the inputs to clear the contents of the field. 
- `app/src/server/db/dbOperations.ts` - modified the `updateDraft` function to take `additions` and `deletions` rather than just one `update` parameter. This seemed like the best approach in order to utilize `$unset` for deletions, but let me know if you have a way that it can be done with just the `update` input.
- `app/src/server/trpc/routers/draft.ts` - updated the input of the `updateDraft` procedure to align with changes in the above two files.

# Testing guidance
- `npm run check:all`
- `npm run start:all`
- Navigate to the authoring tab.
- Create a new Library/Measure through both of the ways (starting from scratch and starting from an existing artifact)
- Make sure that the text inputs are populated correctly if the fields are already defined on the artifact.
- Try deleting/editing the inputs and clicking submit to ensure that they have been updated.
- Do the same but from the route of just viewing the existing artifacts in the local database.
- Make sure changes are only saved when the user clicks "Submit".

Open Questions for the Reviewers:
- When the user deletes the contents of an inputs (for example, `url`), the `url` field is removed from the artifact. If the user then adds something back, then the `url` is added back to the artifact with the new value, but it is added to the bottom of the JSON. Is that okay? 
- Are there any fields that should _always_ exist on the artifact even though it is just a draft? Maybe this is something we want to do later? But for example, should we even make it possible for the user to remove a url from the artifact? When a user creates a new artifact from scratch, the only things that are populated are the url, identifier, and status, so I think this is okay?
- Is there a better way to remove the properties from a Mongo collection? I added the `additions` and `deletions` inputs and used `$set` and `$unset` but wasn't sure if there was a better way.

### DIFF
--- a/app/src/components/ArtifactFieldInput.tsx
+++ b/app/src/components/ArtifactFieldInput.tsx
@@ -1,0 +1,30 @@
+import { ActionIcon, TextInput } from '@mantine/core';
+import { SetStateAction } from 'react';
+import { X } from 'tabler-icons-react';
+
+export interface ArtifactFieldInputProps {
+  label: string;
+  value: string;
+  setField: (value: SetStateAction<string>) => void;
+}
+
+export default function ArtifactFieldInput({ label, value, setField }: ArtifactFieldInputProps) {
+  return (
+    <div>
+      <TextInput
+        label={label}
+        value={value}
+        onChange={e => setField(e.target.value)}
+        rightSection={
+          <ActionIcon
+            onClick={() => {
+              setField('');
+            }}
+          >
+            <X size={16} />
+          </ActionIcon>
+        }
+      />
+    </div>
+  );
+}
--- a/app/src/components/ArtifactFieldInput.tsx
+++ b/app/src/components/ArtifactFieldInput.tsx
@@ -1,11 +1,10 @@
 import { ActionIcon, TextInput } from '@mantine/core';
-import { Dispatch, SetStateAction } from 'react';
 import { X } from 'tabler-icons-react';
 
 export interface ArtifactFieldInputProps {
   label: string;
   value: string;
-  setField: Dispatch<SetStateAction<string>>;
+  setField: (val: string) => void;
 }
 
 export default function ArtifactFieldInput({ label, value, setField }: ArtifactFieldInputProps) {
--- a/app/src/components/ArtifactFieldInput.tsx
+++ b/app/src/components/ArtifactFieldInput.tsx
@@ -4,15 +4,17 @@ import { X } from 'tabler-icons-react';
 export interface ArtifactFieldInputProps {
   label: string;
   value: string;
+  disabled?: boolean;
   setField: (val: string) => void;
 }
 
-export default function ArtifactFieldInput({ label, value, setField }: ArtifactFieldInputProps) {
+export default function ArtifactFieldInput({ label, value, disabled, setField }: ArtifactFieldInputProps) {
   return (
     <div>
       <TextInput
         label={label}
         value={value}
+        disabled={disabled}
         onChange={e => setField(e.target.value)}
         rightSection={
           <ActionIcon
--- a/app/src/components/ArtifactFieldInput.tsx
+++ b/app/src/components/ArtifactFieldInput.tsx
@@ -1,11 +1,11 @@
 import { ActionIcon, TextInput } from '@mantine/core';
-import { SetStateAction } from 'react';
+import { Dispatch, SetStateAction } from 'react';
 import { X } from 'tabler-icons-react';
 
 export interface ArtifactFieldInputProps {
   label: string;
   value: string;
-  setField: (value: SetStateAction<string>) => void;
+  setField: Dispatch<SetStateAction<string>>;
 }
 
 export default function ArtifactFieldInput({ label, value, setField }: ArtifactFieldInputProps) {
--- a/app/src/pages/authoring/[resourceType]/[id].tsx
+++ b/app/src/pages/authoring/[resourceType]/[id].tsx
@@ -126,7 +126,7 @@ export default function ResourceAuthoringPage() {
                     setUrl('');
                   }}
                 >
-                  <X />
+                  <X size={16} />
                 </ActionIcon>
               }
             />
@@ -140,7 +140,7 @@ export default function ResourceAuthoringPage() {
                     setIdentifier('');
                   }}
                 >
-                  <X />
+                  <X size={16} />
                 </ActionIcon>
               }
             />
@@ -154,7 +154,7 @@ export default function ResourceAuthoringPage() {
                     setName('');
                   }}
                 >
-                  <X />
+                  <X size={16} />
                 </ActionIcon>
               }
             />
@@ -168,7 +168,7 @@ export default function ResourceAuthoringPage() {
                     setTitle('');
                   }}
                 >
-                  <X />
+                  <X size={16} />
                 </ActionIcon>
               }
             />
@@ -182,7 +182,7 @@ export default function ResourceAuthoringPage() {
                     setDescription('');
                   }}
                 >
-                  <X />
+                  <X size={16} />
                 </ActionIcon>
               }
             />
--- a/app/src/pages/authoring/[resourceType]/[id].tsx
+++ b/app/src/pages/authoring/[resourceType]/[id].tsx
@@ -13,6 +13,10 @@ export default function ResourceAuthoringPage() {
 
   const [url, setUrl] = useState('');
   const [identifier, setIdentifier] = useState('');
+  const [name, setName] = useState('');
+  const [title, setTitle] = useState('');
+  const [version, setVersion] = useState('');
+  const [description, setDescription] = useState('');
 
   const ctx = trpc.useContext();
 
@@ -58,8 +62,23 @@ export default function ResourceAuthoringPage() {
     }
   });
 
-  function parseUpdate(url: string, identifier: string) {
-    const update: { url?: string; identifier?: fhir4.Identifier[] } = {};
+  function parseUpdate(
+    url: string,
+    identifier: string,
+    name: string,
+    title: string,
+    version: string,
+    description: string
+  ) {
+    const update: {
+      url?: string;
+      identifier?: fhir4.Identifier[];
+      name?: string;
+      title?: string;
+      version?: string;
+      description?: string;
+    } = {};
+
     if (url !== '') {
       update['url'] = url;
     }
@@ -70,6 +89,18 @@ export default function ResourceAuthoringPage() {
       } else {
         update['identifier'] = [{ value: splitIden[0] }];
       }
+    }
+    if (name !== '') {
+      update['name'] = name;
+    }
+    if (title !== '') {
+      update['title'] = title;
+    }
+    if (version !== '') {
+      update['version'] = version;
+    }
+    if (description !== '') {
+      update['description'] = description;
     }
 
     return update;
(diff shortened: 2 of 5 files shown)


app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -58,21 +70,55 @@ export default function ResourceAuthoringPage() {
...
       const splitIden = identifier.split('|');
       if (splitIden.length > 1) {
-        update['identifier'] = [{ system: splitIden[0], value: splitIden[1] }];
+        additions['identifier'] = [{ system: splitIden[0], value: splitIden[1] }];
       } else {
-        update['identifier'] = [{ value: splitIden[0] }];
+        additions['identifier'] = [{ value: splitIden[0] }];
       }
+    } else {
+      deletions['identifier'] = [{ system: '', value: '' }];
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:103]: If the user is allowed to add more than one identifier using `"|"`, I think it might be helpful to emphasize this with the UI. Maybe you could add a tooltip note to the `identifier TextInput` to let the user know how to add multiple types of identifiers. Another idea might be to break this `identifier TextInput` into two text inputs: `A system identifier TextInput `and a `value identifier TextInput`.  Also that way you could display both of the chosen identifiers rather than just one?

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -86,18 +132,87 @@ export default function ResourceAuthoringPage() {
...
+              rightSection={
+                <ActionIcon
+                  onClick={() => {
+                    setUrl('');
+                  }}
+                >
+                  <X />
+                </ActionIcon>
+              }
+            />
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:148]: Since you have a lot of TextInput elements and their layout seems to be the same, would it be helpful to pull this TextInput div into another file and then call it as many times as you need it and pass in the properties as params? This might result in a trade-off I'm not aware of, however.

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -58,21 +70,55 @@ export default function ResourceAuthoringPage() {
...
+      title?: string;
+      description?: string;
+    } = {};
+
     if (url !== '') {
-      update['url'] = url;
+      additions['url'] = url;
+    } else {
+      deletions['url'] = '';
     }
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:94]: This might be an odd point, but if the user were to add a bunch of blank spaces or even just one space (i.e. " ") for an attribute, the attribute's value gets set to the blank space/s. Would it be worthwhile to modify this so that a character of some sort must be entered into the TextInput in order for the attribute's value to be assigned?

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -86,18 +132,87 @@ export default function ResourceAuthoringPage() {
...
+                >
+                  <X />
+                </ActionIcon>
+              }
+            />
             <Button
               w={120}
-              onClick={() =>
+              onClick={() => {
+                const [additions, deletions] = parseUpdate(url, identifier, name, title, description);
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:208]: Just a personal thought, but maybe you could disable the submit button until the user makes direct changes to one of the attributes in a TextInput?  This is more subjective so it really comes down to what you think looks the best :)

BriannaVH (commented):
So far everything is looking great! I love how the UI looks so far.

sarahmcdougall (changes requested):
In addition to in-line comments, I have an additional comment:

Since we are not live-editing the resource content with a code editor, the resource content does not update until the "Submit" button is clicked. I think this is okay, but maybe this should be made more clear to the user. Maybe we could add some kind of header above the Prism component that says something along the lines of "current JSON content" so it's clear that the content is not being updated live

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -86,18 +132,87 @@ export default function ResourceAuthoringPage() {
...
+              label="identifier"
+              value={identifier}
+              onChange={e => setIdentifier(e.target.value)}
+              rightSection={
+                <ActionIcon
+                  onClick={() => {
+                    setIdentifier('');
+                  }}
+                >
+                  <X />
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:159]: LOL yeah good call!

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:94]: Good catch & idea!

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:103]: So I think that this is definitely something we are going to look into in the future, but for now I think it's out of scope for this task since this is how we already had it. See the FHIR spec [here](http://hl7.org/fhir/R4/datatypes.html#identifier) to see how identifier is defined since it's technically not more than one identifier. No need to worry too much about this right now though :)

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:208]: I like this idea! I am going to see if there is a not-too-complicated way to do this.

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:148]: Great suggestion! I was able to pull this into a file in the components directory called `ArtifactFieldInput.tsx`

BriannaVH (approved):
Everything looks great! I think the UI looks really good and is intuitive! :)

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -116,76 +117,11 @@ export default function ResourceAuthoringPage() {
...
-                <ActionIcon
-                  onClick={() => {
-                    setDescription('');
-                  }}
-                >
-                  <X size={16} />
-                </ActionIcon>
-              }
-            />
+            <ArtifactFieldInput label="url" value={url} setField={setUrl} />
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:140]: Shoutout @BriannaVH 😄

app/src/components/ArtifactFieldInput.tsx
```diff
@@ -0,0 +1,30 @@
+import { ActionIcon, TextInput } from '@mantine/core';
+import { SetStateAction } from 'react';
+import { X } from 'tabler-icons-react';
+
+export interface ArtifactFieldInputProps {
+  label: string;
+  value: string;
+  setField: (value: SetStateAction<string>) => void;
```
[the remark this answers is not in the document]
elsaperelli: Gotcha! I honestly just did what I had based on what the type was, but this seems like a better solution.

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (mgramigna) [reply app/src/pages/authoring/[resourceType]/[id].tsx:120]: I still think this input could use an update though, at least with a tooltip to say that `system|version` is supported for a single identifier. I'd say it's in-scope of this task to update the text field to indicate supported formats

app/src/components/ArtifactFieldInput.tsx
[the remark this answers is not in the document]
mgramigna: I usually just use `(val: string) => void;` instead of `Dispatch<SetStateAction>` since it's cleaner/is assignable to that type

elsaperelli:
> In addition to in-line comments, I have an additional comment:
> 
> Since we are not live-editing the resource content with a code editor, the resource content does not update until the "Submit" button is clicked. I think this is okay, but maybe this should be made more clear to the user. Maybe we could add some kind of header above the Prism component that says something along the lines of "current JSON content" so it's clear that the content is not being updated live

Ah I forgot to do this! This is what I have for this right now, let me know what you think.

image

sarahmcdougall:
> Ah I forgot to do this! This is what I have for this right now, let me know what you think.

Looks good to me! 👍

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:120]: Gotcha! I spoke to Matt and we decided it made the most sense to just have the identifier value and the identifier system be different inputs. The identifier system is optional, so it is only saved on the artifact JSON if the identifier value exists.

sarahmcdougall (changes requested):

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -31,11 +77,26 @@ export default function ResourceAuthoringPage() {
...
+        setIdentifierValue(cmsIdentifier.value);
+        if (cmsIdentifier.system) {
+          setIdentifierSystem(cmsIdentifier.system);
+        }
       } else if (resource.identifier[0].value) {
-        setIdentifier(resource.identifier[0].value);
+        setIdentifierValue(resource.identifier[0].value);
+        if (resource.identifier[0].system) {
+          setIdentifierSystem(resource.identifier[0].value);
+        }
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:88]: I've run into an issue where if you go to edit a resource without a value or system identifier and add both of them to the resource, the `system identifier` gets overwritten with the `value identifier`. Also, if you delete the value identifier but don't delete the system identifier, the identifier will be deleted from the json entirely.

BriannaVH (changes requested):
Other than the error having to do with the identifiers, everything is looking really good and I think it's great how much you've been able to reduce the amount of code written :)

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -31,11 +77,26 @@ export default function ResourceAuthoringPage() {
...
-        setIdentifier(cmsIdentifier.value);
+        setIdentifierValue(cmsIdentifier.value);
+        if (cmsIdentifier.system) {
+          setIdentifierSystem(cmsIdentifier.system);
+        }
       } else if (resource.identifier[0].value) {
-        setIdentifier(resource.identifier[0].value);
+        setIdentifierValue(resource.identifier[0].value);
+        if (resource.identifier[0].system) {
+          setIdentifierSystem(resource.identifier[0].value);
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:87]: Another good catch!

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx]: Nice catch with the overwritten stuff!! For the other part of the comment, I believe that is actually the desired functionality since we just don't want the user to be able to define only an identifier system- the user should only be able to enter one of three things:
- nothing for either identifier.system or identifier.value
- just identifier.value
- both identifier.system and identifier.value

BriannaVH (approved):
Everything looks great to me!

sarahmcdougall (approved):

BriannaVH (approved):
Everything looks great!

PR #45, opened 2023-06-27
Add button for deleting draft artifacts

sarahmcdougall:
# Summary
Adds button on each draft resource card to delete it. When clicked, the draft artifact is removed from the database and is no longer reflected via the draft resource counts or the rendered resource cards.

## New behavior
For draft resources, the user now has the option to delete a draft resource from the UI. When the trash icon is clicked, a confirmation modal appears to confirm the deletion. When confirmed, the resource is deleted from the database, its resource card is no longer rendered, and the draft artifact count is refreshed to reflect the updated count.

## Code changes
* `server/trpc/routers/draft.ts` - add new trpc procedure for deleting a draft artifact
* `server/db/dbOperations.ts` - add new function to delete the selected resource from the mongo collection
* `components/ResourceInfoCard.tsx`
    * Add delete button to each card, for authoring only, that pulls up a confirmation modal when clicked
    * Adds deletion confirmation modal that appears when the trash icon is clicked (probably easier to view this change locally rather than looking at the Github code diff)
    * When confirmed, a success notification appears, the resource is deleted from the database, and the draft counts and rendered resource cards are refreshed. If an error occurs, an error notification appears
* `components/ConfirmationModal.tsx` - new confirmation modal component, designed to be reusable if we have other confirmation modal uses in the future
* `util/types/fhir.ts` - update the resource card’s `resourceType` type to be `ArtifactResourceType` instead of `string` (since the rendered resource cards will always be of type Library or Measure)

# Testing guidance
* Navigate to the authoring mode and click on one of the resource count buttons to bring you to the resource cards for a given resource type. Check that the trash icon is rendered on each card
* Click on the trash icon. Check that the confirmation modal appears
* Click on “Delete” from the confirmation modal. Check the following:
    * The resource’s card no longer appears on the page
    * The draft artifact button for the given resource type is decremented by 1
    * The artifact is not in the database. You can do this by viewing the resource type collection in Robo3T
* Go through the same process, but instead click on “Cancel” from the confirmation modal. Check that the resource has not been deleted, and the UI has not changed in any way.
* Navigate to the repository mode, and check that the service resource cards have not been altered in any way (i.e. they should not have the trash icon, and they should still have the “details” icon rather than the “edit” icon)


app/src/components/ResourceInfoCard.tsx
```diff
@@ -1,8 +1,22 @@
...
+  ActionIcon,
+  Grid,
+  Paper,
+  Text,
+  createStyles,
+  em,
+  getBreakpointValue,
+  rem,
+  Tooltip,
+  Group
```
BriannaVH, line 11: I know you mentioned in the past we should add imports in alphabetical order. This might not apply since it is a different file but just thought I should mention it.

app/src/components/ResourceInfoCard.tsx
```diff
@@ -22,50 +36,105 @@ const useStyles = createStyles(theme => ({
...
-            </ActionIcon>
-          </Tooltip>
-        </Link>
-      </Grid>
-    </Paper>
+    <>
+      <ConfirmationModal
+        open={isConfirmationModalOpen}
+        onClose={() => setIsConfirmationModalOpen(false)}
+        title={'Delete Draft?'}
```
BriannaVH, line 68: This is totally subjective but I'm not a huge fan of the title in the delete confirmation modal. I think the modal text suffices for letting the user know what to do. I edited it and included the screenshot to show what it would look like without the title included! 

Screenshot 2023-06-27 at 1 34 48 PM

BriannaVH (approved):
Overall, I think things look great. I didn't note much to change other than a few minor things that you can choose to include or not. Great Job!

app/src/server/trpc/routers/draft.ts
```diff
@@ -36,5 +43,11 @@ export const draftRouter = router({
     .input(z.object({ resourceType: z.enum(['Measure', 'Library']), draft: z.any(), id: z.string() }))
     .mutation(async ({ input }) => {
       return updateDraft(input.resourceType, input.id, input.draft);
+    }),
+
+  deleteDraft: publicProcedure
+    .input(z.object({ id: z.string().optional(), resourceType: z.enum(['Measure', 'Library']).optional() }))
```
lmd59, line 49: I'm still getting up to speed on trpc, but is there a reason why we set both properties (id and resourceType) to optional, but then immediately check they exist before deleting? Should we just not mark them as optional?

sarahmcdougall: After doing some more testing here, it doesn't seem like we need to make the inputs optional as I originally did here.
In general, some of the draft procedures throw TRPC client errors when the page first loads because the FHIR resource type will be undefined on first load. To avoid this TRPC error, procedures like `getDrafts` and `getDraftById` have optional inputs that get checked before executing the query. Since `deleteDraft` does not get executed on first load (it's a mutation that gets triggered by a button click), I am not getting any TRPC client errors when the inputs are required instead of optional. I'll update this!

app/src/components/ResourceInfoCard.tsx
sarahmcdougall: I worked with this a bit and I think the modal looks best without the title since this the action is pretty simple here (deleting a resource) - I removed the title for now

app/src/components/ResourceInfoCard.tsx
```diff
@@ -22,50 +36,104 @@ const useStyles = createStyles(theme => ({
...
-          </Tooltip>
-        </Link>
-      </Grid>
-    </Paper>
+    <>
+      <DeletionConfirmationModal
+        open={isConfirmationModalOpen}
+        onClose={() => setIsConfirmationModalOpen(false)}
+        modalText={`This will delete draft ${resourceInfo.resourceType} "${
+          resourceInfo.name ? resourceInfo.name : `${resourceInfo.resourceType}/${resourceInfo.id}`
```
BriannaVH, line 69: Awesome! I think the Modal Component looks great without the title!

BriannaVH (approved):
Looks great! I think you're all set to go!

lmd59 (approved):
Looks good!

PR #46, opened 2023-06-27
Make Data Requirements Human Readable

BriannaVH:
# Summary
This PR allows users to toggle between two different views for the data requirements of a specified resource: The raw version of the data requirements and a readable version of the data requirements array.

## New behavior
The data requirements panel now populates with a `Segment Controller` mantine component at the top of the panel where the user can manually toggle between two different displays of the data requirements. The panel should always initially start out on the raw data requirements view. If the user manually selects the " readable data requirements" view they will then see a series of data requirement cards. Each card has an icon in the upper right hand corner the user can click to expand/constrict the card. Each card should display specific information of one element from the data requirements array. Specifically, it's fields should include any of it's `code filters, dateFilters, extensions, and types`. There are some resources for which the raw data requirements should appear but the readable version shouldn’t because the data requirements array doesn’t have any elements. In these instances, the Segment Control component which allows the user to slide between different views should also not appear.

## Code changes
` <SegmentedControl>` —> I use a segmented control component in `app/src/pages/repository/[resourceType]/[id].tsx` to allow the user to choose between the two different views of the data requirements
`app/src/components/DataRequirements.tsx` —> Extracts the different data requirements fields and displays them in a card component that it returns
`Accordion` --> The Accordion component in `app/src/components/DataRequirements.tsx` is what allows the cards to be collapsible upon a user's click.

# Testing guidance
```
 npm run check:all
 npm run start:all
```
Navigate to the repository tab and select a library or measure resource. Then request data requirements for the resource. If the data requirements exist the data requirement's tab/panel should appear populated with a segment controller component at the top. The user should initially see the raw data requirements and it should look like the following:
Screenshot 2023-06-27 at 2 21 09 PM

Following this, toggle the segment controller so that "Readable Data Requirements" is now selected and the cards should appear looking like the following: 
Screenshot 2023-06-27 at 4 20 53 PM

To make sure they are correct, you can toggle back and forth between the two views to ensure that each of the element's in the data requirements array accurately reflects what each of the card data requirements are displaying.In the raw data requirements tab, the data requirements array should have the exact same number of elements as the readable version has cards. You can also compare the contents of both the cards and each element of the data requirements array to make sure they match, and this should be made easier given that they both are in the same order.  Some resources (`Library/CumulativeMedicationDurationFHIR4` for instance) should not have readable data requirements, even if they have the raw data requirements. In instances like these, the Segment Controller should not appear for the user to toggle between them.

### DIFF
--- a/app/src/components/DataRequirements.tsx
+++ b/app/src/components/DataRequirements.tsx
@@ -54,12 +54,12 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
                 <List withPadding>
                   {extension?.map(e => (
                     <>
-                      <List.Item>
+                      <List.Item key={e.url}>
                         <b>URL:</b>
                         {e.url}
                       </List.Item>
-                      {e?.valueString && (
-                        <List.Item>
+                      {e.valueString && (
+                        <List.Item key={e.valueString}>
                           <b> ValueString: </b>
                           {e.valueString}
                         </List.Item>
@@ -74,14 +74,14 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
                       <b>DateFilter(s):</b>
                     </List.Item>
                     <List withPadding>
-                      {dateFilter?.map(data => (
+                      {dateFilter.map(date => (
                         <>
-                          <List.Item>
+                          <List.Item key={date?.valuePeriod?.start}>
                             <b> Start Date: </b>
-                            {data?.valuePeriod?.start} <br />
+                            {date?.valuePeriod?.start} <br />
                           </List.Item>
-                          <List.Item>
-                            <b> End Date: </b> {data?.valuePeriod?.end} <br />
+                          <List.Item key={date?.valuePeriod?.end}>
+                            <b> End Date: </b> {date?.valuePeriod?.end} <br />
                           </List.Item>
                         </>
                       ))}
@@ -101,12 +101,12 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
                       {codes.map(codeInterface => (
                         <>
                           {codeInterface.code && (
-                            <List.Item>
-                              <b> Code: </b> {codeInterface?.code} <br />
+                            <List.Item key={codeInterface.code}>
+                              <b> Code: </b> {codeInterface.code} <br />
                             </List.Item>
                           )}
                           {codeInterface.system && (
-                            <List.Item>
+                            <List.Item key={codeInterface.system}>
                               <b> System: </b> {codeInterface?.system} <br />
                             </List.Item>
                           )}
@@ -122,8 +122,8 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
                       <b>Value Set(s):</b>
                     </List.Item>
                     <List withPadding>
-                      {codeFiltersWithVS.map(c => (
-                        <List.Item>
+                      {codeFiltersWithVS.map((c, index) => (
+                        <List.Item key={index}>
                           {c.valueSet} <br />
                         </List.Item>
                       ))}
--- a/app/src/components/DataRequirements.tsx
+++ b/app/src/components/DataRequirements.tsx
@@ -1,6 +1,6 @@
 import { Accordion, Center, createStyles, em, getBreakpointValue, List, Paper, rem } from '@mantine/core';
 
-interface codeInterface {
+interface CodeInterface {
   code: string | undefined;
   system: string | undefined;
 }
@@ -22,15 +22,15 @@ const useStyles = createStyles(theme => ({
  */
 function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.DataRequirement) {
   const { classes } = useStyles();
-  const codes: codeInterface[] = [];
+  const codes: CodeInterface[] = [];
 
-  const valueSets = codeFilter?.filter(function (x: fhir4.DataRequirementCodeFilter) {
-    return x?.valueSet !== undefined;
+  const codeFiltersWithVS = codeFilter?.filter(cf => {
+    return cf?.valueSet !== undefined;
   });
 
-  codeFilter?.forEach((arr: fhir4.DataRequirementCodeFilter) => {
-    arr?.code?.forEach((element: fhir4.Coding) => {
-      const code: codeInterface = { code: element?.code, system: element?.system };
+  codeFilter?.forEach(cf => {
+    cf?.code?.forEach(c => {
+      const code: CodeInterface = { code: c.code, system: c.system };
       codes.push(code);
     });
   });
@@ -41,7 +41,6 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
         <Accordion variant="contained" radius="lg" defaultValue="customization">
           <Accordion.Item value="customization">
             <Accordion.Control>
-              {' '}
               <Center>
                 <h2>{type} </h2>
               </Center>
@@ -53,18 +52,16 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
                   <br />
                 </List.Item>
                 <List withPadding>
-                  {extension?.map((item: any, index: any, index2: any) => (
+                  {extension?.map(e => (
                     <>
-                      {item?.url != undefined && (
-                        <List.Item key={index}>
-                          <b> URL: </b>
-                          {item.url}{' '}
-                        </List.Item>
-                      )}
-                      {item?.valueString != undefined && (
-                        <List.Item key={index2}>
+                      <List.Item>
+                        <b>URL:</b>
+                        {e.url}
+                      </List.Item>
+                      {e?.valueString && (
+                        <List.Item>
                           <b> ValueString: </b>
-                          {item.valueString}{' '}
+                          {e.valueString}
                         </List.Item>
                       )}
                       <br />
@@ -77,14 +74,14 @@ function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.Dat
                       <b>DateFilter(s):</b>
                     </List.Item>
                     <List withPadding>
-                      {dateFilter?.map((item: fhir4.DataRequirementDateFilter, index: any, index2: any) => (
+                      {dateFilter?.map(data => (
                         <>
-                          <List.Item key={index}>
+                          <List.Item>
                             <b> Start Date: </b>
-                            {item?.valuePeriod?.start} <br />
+                            {data?.valuePeriod?.start} <br />
                           </List.Item>
-                          <List.Item key={index2}>
-                            <b> End Date: </b> {item?.valuePeriod?.end} <br />
+                          <List.Item>
+                            <b> End Date: </b> {data?.valuePeriod?.end} <br />
                           </List.Item>
                         </>
                       ))}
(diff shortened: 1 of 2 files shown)


app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,159 @@
...
+                          <List.Item key={index}>
+                            <b> Start Date: </b>
+                            {new Date(item.valuePeriod.start).getUTCMonth() + 1}/
+                            {new Date(item.valuePeriod.start).getUTCDate()}/
+                            {new Date(item.valuePeriod.start).getUTCFullYear()} <br />
+                          </List.Item>
+                          <List.Item key={index2}>
+                            <b> End Date: </b> {new Date(item.valuePeriod.end).getUTCMonth() + 1}/
+                            {new Date(item.valuePeriod.end).getUTCDate()}/
+                            {new Date(item.valuePeriod.end).getUTCFullYear()} <br />
```
sarahmcdougall, line 101: I like the look of displaying the period start/end with this format, but since the period start/end can be a dateTime, it might be best to just display `item.valuePeriod.start` and `item.valuePeriod.end` as is, without any sort of conversion. That way, we will still have the millisecond precision. This isn't super important for the sake of this demo application, but it might be nice to have this precision so that our precision is consistent across all of our applications (for example, the fqm-execution library uses millisecond precision for date periods).

sarahmcdougall (changes requested):
Amazing work on this, especially considering all the unknowns for how we wanted to display the data requirements! The human-readable format is a lot easier to digest compared to the JSON that we get back from the server.

Let me know if you have any questions regarding my suggestions - most of them are TypeScript-specific. As mentioned in one of the in-line comments, I expect that some of the TypeScript type changes will lead to other changes, as other parts of the code will be affected by the type changes. Happy to pair program on those changes if you would like :)

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -158,11 +160,31 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+          {dataRequirements?.resourceType === 'Library' && dataRequirements?.dataRequirement != undefined && (
             <Tabs.Panel value="data-requirements">
-              <Prism language="json" colorScheme="light">
-                {JSON.stringify(dataRequirements, null, 2)}
-              </Prism>
+              <Center>
+                {dataRequirements?.dataRequirement.length > 0 && (
+                  <SegmentedControl
+                    fullWidth
+                    color="blue"
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx:169]: Yes, I also agree it looks better with it's default color!

app/src/components/DataRequirements.tsx
BriannaVH: I agree it would be more precise, but I've noticed when I don't convert the time objects, it only ever shows 0's for the milliseconds. This screenshot is an example of what they all look like: 
Screenshot 2023-06-29 at 12 29 08 PM

Are they all supposed to have 0's or is it not working as intended?

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,145 @@
...
+                          {item?.code && (
+                            <List.Item key={index}>
+                              <b> Code: </b> {item?.code} <br />
+                            </List.Item>
+                          )}
+                          {item?.system && (
+                            <List.Item key={index2}>
+                              <b> System: </b> {item?.system} <br />
+                            </List.Item>
+                          )}
```
sarahmcdougall, line 115: Similar situation here to what I suggested for the extension rendering:
1. Based on the way we gathered `codes`, we know that `codes` will always be defined, especially since we do a check to make sure its length is > 0. So we don't need to use the `?` operator on `codes` or on `item`.
2. `item` will be of type `codeInterface` (since `codes` is an array of `codeInterface`s). So we can update the type for `item`
Good call checking that `code` and `system` exist before rendering them in the list :)

```suggestion
                      {codes.map((item: codeInterface, index: any, index2: any) => (
                        <>
                          {item.code && (
                            <List.Item key={index}>
                              <b> Code: </b> {item.code} <br />
                            </List.Item>
                          )}
                          {item.system && (
                            <List.Item key={index2}>
                              <b> System: </b> {item.system} <br />
                            </List.Item>
                          )}
```

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,145 @@
...
+                  <List withPadding>
+                    <List.Item>
+                      <b>Value Set(s):</b>
+                    </List.Item>
+                    <List withPadding>
+                      {valueSets?.map((item: any, index3: any) => (
+                        <List.Item key={index3}>
+                          {item?.valueSet} <br />
+                        </List.Item>
+                      ))}
```
sarahmcdougall, line 132:
```suggestion
                      {valueSets.map((item: fhir4.DataRequirementCodeFilter, index3: any) => (
                        <List.Item key={index3}>
                          {item.valueSet} <br />
                        </List.Item>
                      ))}
```
Same thing as above - we checked that `valueSets` is defined, so no need for the `?` operator, and we can also change the type of `item` to be `fhir4.DataRequirementCodeFilter` since `valueSets` is of type `fhir4.DataRequirementCodeFilter[]`

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,145 @@
+import { Accordion, Center, createStyles, em, getBreakpointValue, List, Paper, rem } from '@mantine/core';
+
+interface codeInterface {
```
sarahmcdougall, line 3:
```suggestion
interface CodeInterface {
```
Nitpick, but it's best practice to capitalized interfaces in TypeScript - changing this will cause a few resulting changes throughout this file anywhere that `codeInterface` is used

app/src/components/DataRequirements.tsx
sarahmcdougall: If you compare the start date and end date that are shown here to the `valuePeriod` start and end on the raw data requirements, they should match up. My guess is that for this data requirement, yes the precision should be all 0's because that is what is listed on the data requirement.

Example:
Readable data requirements:
Screenshot 2023-06-30 at 11 13 11 AM

Corresponding value period:
Screenshot 2023-06-30 at 11 13 32 AM

(above example is for the first Encounter data requirement for the ColorectalCancerScreeningsFHIR measure)

sarahmcdougall (changes requested):
Great job implementing the changes I suggested :) I have a few more code suggestions - let me know if you have any questions or want to discuss anything further!

Looks like this branch also needs a rebase - I think the rebase will capture the changes I made for adding the delete button for draft artifacts. So fingers crossed that the rebase is fairly simple since I didn't edit these files for creating the deletion functionality

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -158,11 +160,36 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
-              </Prism>
+              <Center>
+                {dataRequirements?.dataRequirement.length > 0 && (
+                  <SegmentedControl
+                    fullWidth
+                    value={dataReqsView}
+                    onChange={setDataReqsView}
+                    data={[
+                      { label: 'Raw Data Requirements', value: 'raw' },
+                      { label: 'Readable Data Requirements', value: 'readable' }
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: I agree it sounds better!

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -158,11 +160,36 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+                </Prism>
+              )}
+              {dataReqsView === 'readable' &&
+                dataRequirements?.dataRequirement.map((item: fhir4.DataRequirement, index: any) => (
+                  <DataReqs
+                    key={index}
+                    type={item?.type}
+                    codeFilter={item?.codeFilter}
+                    dateFilter={item?.dateFilter}
+                    extension={item?.extension}
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: Thanks for explaining, it is definitely a bit confusing but your explanation made it much easier to understand!

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -158,11 +160,36 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
           )}
-          {dataRequirements?.resourceType === 'Library' && (
+          {dataRequirements?.resourceType === 'Library' && dataRequirements?.dataRequirement && (
             <Tabs.Panel value="data-requirements">
-              <Prism language="json" colorScheme="light">
-                {JSON.stringify(dataRequirements, null, 2)}
-              </Prism>
+              <Center>
+                {dataRequirements?.dataRequirement.length > 0 && (
+                  <SegmentedControl
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx]: Totally agree, looks a bunch better with the spacing imo

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,145 @@
...
+                <List withPadding>
+                  {extension?.map((item: any, index: any, index2: any) => (
+                    <>
+                      {item?.url != undefined && (
+                        <List.Item key={index}>
+                          <b> URL: </b>
+                          {item.url}{' '}
+                        </List.Item>
+                      )}
+                      {item?.valueString != undefined && (
```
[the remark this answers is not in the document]
BriannaVH: Thank you for the advice ! I actually ended up going a different route with regard to the mapping so I hopefully don't even need to declare the datatypes at all!

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,142 @@
...
+ * Component which displays all data requirements of a specified resource and displays them
+ * as resource cards that contain all required information
+ */
+function DataRequirements({ type, extension, dateFilter, codeFilter }: fhir4.DataRequirement) {
+  const { classes } = useStyles();
+  const codes: CodeInterface[] = [];
+
+  const codeFiltersWithVS = codeFilter?.filter(cf => {
+    return cf?.valueSet !== undefined;
+  });
```
elsaperelli, line 29:
```suggestion
  const codeFiltersWithVS = codeFilter?.filter(cf => cf?.valueSet);
```
You can shorten this a bit!

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    </List.Item>
+                    <List withPadding>
+                      {dateFilter.map(date => (
+                        <>
+                          <List.Item key={date?.valuePeriod?.start}>
+                            <b> Start Date: </b>
+                            {date?.valuePeriod?.start} <br />
+                          </List.Item>
+                          <List.Item key={date?.valuePeriod?.end}>
+                            <b> End Date: </b> {date?.valuePeriod?.end} <br />
```
elsaperelli, line 84:
```suggestion
                          <List.Item key={date.valuePeriod?.start}>
                            <b> Start Date: </b>
                            {date.valuePeriod?.start} <br />
                          </List.Item>
                          <List.Item key={date.valuePeriod?.end}>
                            <b> End Date: </b> {date.valuePeriod?.end} <br />
```
I don't think we need these since we are doing a `.map` on the dateFilter array so we know each entry is defined as a `DataRequirementDateFilter`

app/src/components/DataRequirements.tsx
```diff
@@ -0,0 +1,142 @@
...
+                      {codes.map(codeInterface => (
+                        <>
+                          {codeInterface.code && (
+                            <List.Item key={codeInterface.code}>
+                              <b> Code: </b> {codeInterface.code} <br />
+                            </List.Item>
+                          )}
+                          {codeInterface.system && (
+                            <List.Item key={codeInterface.system}>
+                              <b> System: </b> {codeInterface?.system} <br />
```
elsaperelli, line 110:
```suggestion
                              <b> System: </b> {codeInterface.system} <br />
```

elsaperelli (changes requested):

sarahmcdougall (approved):
Great changes!!

elsaperelli (commented):
Another thing I just noticed is that when I pull put the data requirements in a smaller window and then I resize that window, the data requirements stay small! I haven't looked into this a bunch yet, but I imagine it has something to do with the resize window/useEffect :) 

image

elsaperelli (approved):
I just realized that the scroll bar/window resize issue also occurs in `ResourceCards.tsx`...which was from me! Oops haha. So don't worry about this for your task- we can add a backlog task or save it as an end of sprint activity :) Good job with this!

PR #47, opened 2023-07-03
Render Resource Dependencies

BriannaVH:
# Summary
The user can now render the dependencies/relatedArtifacts of a resource!

## New behavior
By clicking the "Get Data Requirements" button for a resource, a new `"Dependencies" mantine tab` will pop up. Each of the resource's dependencies and their respective fields  should display in a card component. If the dependency is another library/measure resource then a `SquareArrowRight icon` should appear on the card that, when clicked, links to that resource's own page. 

## Code changes
`app/src/components/DependencyCards.tsx` —> Extracts the different fields and properties of each of the dependencies of a resource and displays them in a card component
` ScrollArea` —> The scrollArea component allows users to scroll through a resource's dependencies and continue to see the top of the page
`Link `--> The user can click the `SquareArrowRight icon` on dependencies to navigate to the resource's own page using the link component.

# Testing guidance
```
 npm run check:all
 npm run start:all
```
Navigate to the repository tab and select a library or measure resource. Then request data requirements for the resource. If the data requirements exist the Dependency tab/panel should appear. The panel should contain card components for each of the resource's dependencies. Every card should display the name and link of the dependency. If the dependency is a Library or Measure resource, the  `SquareArrowRight icon` should appear on the card and should link to the page of the resource when clicked. To make sure the dependencies are being properly displayed, you can navigate to the related Artifacts section in the data requirements tab. Each of the related Artifacts should be displayed accurately.

### DIFF
--- a/app/src/components/DependencyCards.tsx
+++ b/app/src/components/DependencyCards.tsx
@@ -47,9 +47,10 @@ function Dependencies(props: myComponentProps) {
   let dependencyInfo: DependencyInformation = {};
 
   // TODO
-  //Theoretically, canonicals for FHIR resources should be formatted in a way such that this code will work
-  //to get the proper url to route to. However, this might not always be the case and so we should add
-  //some logic to to handle trying to view resource details using the canoncical URL fo the resource as
+  //Canonicals for FHIR resources should be formatted in such a way that this code will work
+  //to get the proper url that routes the user to the new page.
+  // However, they can technically be anything and so we should potentially add
+  //some logic that handles viewing resource details using the canoncical URL of the resource as
   //the identifying information
   if (resourceLink?.includes('Library')) {
     const resourceArr: string[] = resourceLink.substring(resourceLink.indexOf('Library')).split('|');
--- a/app/src/components/DependencyCards.tsx
+++ b/app/src/components/DependencyCards.tsx
@@ -1,0 +1,84 @@
+import {
+  ActionIcon,
+  Center,
+  createStyles,
+  em,
+  Grid,
+  Paper,
+  Text,
+  Tooltip,
+  getBreakpointValue,
+  rem
+} from '@mantine/core';
+import Link from 'next/link';
+import React from 'react';
+import { SquareArrowRight } from 'tabler-icons-react';
+
+export interface ResourceInformation {
+  type?: string;
+  link?: string;
+}
+
+const useStyles = createStyles(theme => ({
+  card: {
+    borderRadius: 6,
+    border: `${rem(2)} solid ${theme.colors.gray[3]}`,
+    width: '800px',
+    [`@media (max-width: ${em(getBreakpointValue(theme.breakpoints.lg) - 1)})`]: {
+      width: '100%'
+    }
+  }
+}));
+function Dependencies(props: { value: fhir4.RelatedArtifact; sourceName: string | undefined }) {
+  const { classes } = useStyles();
+  const type = props.value.type;
+  const display = props.value.display;
+  const resource = props.value.resource;
+  const sourceName = JSON.stringify(props.sourceName);
+
+  let resourceInfo: ResourceInformation = {};
+
+  if (resource?.includes('Library')) {
+    let substr = resource.substring(resource.indexOf('Library'));
+    let myArray: string[] = substr?.split('|');
+    resourceInfo = { type: 'Library', link: myArray[0] };
+  } else if (resource?.includes('Measure')) {
+    let substr = resource.substring(resource.indexOf('Measure'));
+    let myArray: string[] = substr?.split('|');
+    resourceInfo = { type: 'Measure', link: myArray[0] };
+  }
+
+  return (
+    <>
+      <Center>
+        <Paper className={classes.card} shadow="sm" p="md">
+          <Grid align="center">
+            <Grid.Col span={10}>
+              <div>
+                <Text size="lg" fw={700}>
+                  {display}
+                </Text>
+              </div>
+              <div>
+                <Text size="sm" fw={500}>
+                  {resource}
+                </Text>
+              </div>
+            </Grid.Col>
+            {resourceInfo?.type && (
+              <Link href={`http://localhost:3001/${resourceInfo?.link}#JSON`} key={1}>
+                <Tooltip label={'Open to Resource'} openDelay={1000}>
+                  <ActionIcon radius="md" size="md" variant="subtle" color="gray">
+                    {<SquareArrowRight size="24" />}
+                  </ActionIcon>
+                </Tooltip>
+              </Link>
+            )}
+          </Grid>
+        </Paper>
+      </Center>
+    </>
+  );
+}
+
+export default Dependencies;
--- a/app/src/components/DependencyCards.tsx
+++ b/app/src/components/DependencyCards.tsx
@@ -1,5 +1,4 @@
 import {
-  Anchor,
   ActionIcon,
   Center,
   createStyles,
@@ -11,6 +10,7 @@ import {
   getBreakpointValue,
   rem
 } from '@mantine/core';
+import Link from 'next/link';
 import React from 'react';
 import { SquareArrowRight } from 'tabler-icons-react';
 
@@ -68,13 +68,13 @@ function Dependencies(props: { relatedArtifact: fhir4.RelatedArtifact }) {
               </div>
             </Grid.Col>
             {dependencyInfo.type && (
-              <Anchor href={`/${dependencyInfo.link}`}>
+              <Link href={`/${dependencyInfo.link}`}>
                 <Tooltip label={'Open Dependency Resource'} openDelay={1000}>
                   <ActionIcon radius="md" size="md" variant="subtle" color="gray">
                     {<SquareArrowRight size="24" />}
                   </ActionIcon>
                 </Tooltip>
-              </Anchor>
+              </Link>
             )}
           </Grid>
         </Paper>
--- a/app/src/components/DependencyCards.tsx
+++ b/app/src/components/DependencyCards.tsx
@@ -1,4 +1,5 @@
 import {
+  Anchor,
   ActionIcon,
   Center,
   createStyles,
@@ -10,11 +11,10 @@ import {
   getBreakpointValue,
   rem
 } from '@mantine/core';
-import Link from 'next/link';
 import React from 'react';
 import { SquareArrowRight } from 'tabler-icons-react';
 
-export interface ResourceInformation {
+interface DependencyInformation {
   type?: string;
   link?: string;
 }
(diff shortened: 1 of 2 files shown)


lmd59 (commented):
A couple of questions and considerations!

- It looks like we're not saving either dependencies or data requirements to storage (i.e. does not retain on reload). Should it?
- Consider ordering the list or grouping valuesets and codesystems in the dependencies list.
- Could `DependencyCards` reuse the existing card structure (i.e. `ResourceInfoCards`)?
- Should we rename "getDataRequirements" button since it's now creating two tabs, (not just the "Data Requirements" tab? 
Interested in your thoughts on these!

BriannaVH:
> A couple of questions and considerations!
> 
> * It looks like we're not saving either dependencies or data requirements to storage (i.e. does not retain on reload). Should it?
> * Consider ordering the list or grouping valuesets and codesystems in the dependencies list.
> * Could `DependencyCards` reuse the existing card structure (i.e. `ResourceInfoCards`)?
> * Should we rename "getDataRequirements" button since it's now creating two tabs, (not just the "Data Requirements" tab?
>   Interested in your thoughts on these!

I think those are some great recommendations, thank you! I'm temporarily holding off on pushing any other commits on this branch until my other task is merged and I can rebase this one properly. 

To respond to your suggestions,
- For your first recommendation, I totally agree! I think it's much better to have the dependencies stored when the user navigates to a different artifact via a dependency card. It doesn't make much sense that the user would have to request the data requirements of an artifact more than once simply because they clicked a link. I think the functionality is a lot more intuitive with the changes implemented
- I also think your second suggestion where I order the list based on type is a great idea! The user can more easily navigate the dependencies that way 
- As for reusing the resourceInfoCards, I considered that in the start because it would reduce the code a ton, but I think that resourceInfoCard's functionality differs a bit too much to be able to make them work together. I'm totally open to disagreements however!
- I think the "getDataRequirements" button name should definitely change because its getting the dependencies now as well, but I am currently considering  entirely getting rid of the get data requirements button after my other task is merged.

mgramigna:
It might be nice to have a way to look for certain dependencies without having to scroll. Thoughts on maybe playing around with a search box or something where the user can type in a query that filters the list to items containing that query's content in either the display value or url of the dependency?

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -211,6 +245,21 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+          {dataRequirements?.relatedArtifact && (
+            <Tabs.Panel value="dependencies">
+              <Space h="md" />
+              <Text c="dimmed">
+                Number of Dependencies:<b> {dataRequirements.relatedArtifact.length} </b>
+              </Text>
+              <Space h="md" />
+              <ScrollArea.Autosize mah={height * 0.8} type="hover">
+                {dataRequirements.relatedArtifact.map((relatedArtifact, index) => (
+                  <Dependency key={index} relatedArtifact={relatedArtifact} />
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:257]: This is super niche, but it is actually not recommended by React to use the index of iteration as the key on a DOM element. The docs list a pitfall for this [here](https://react.dev/learn/rendering-lists#why-does-react-need-keys). In short, it could cause weird bugs or performance issues in some cases.

Instead, we should make or use a string that is unique for each dependency. I think it's safe to assume that for a given `relatedArtifact`, the `.resource` attribute will be unique. So you can probably use `relatedArtifact.resource` as the key.

If this wasn't the case, the following strategy also works:

```tsx
                {dataRequirements.relatedArtifact.map((relatedArtifact, index) => {
                  // Using the index outside of the actual `key` prop assignment is an okay workaround
                  const key = `dependency-${index}`;
                  return <Dependency key={key} relatedArtifact={relatedArtifact} />;
                })}
```

app/src/components/DependencyCards.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+}));
+
+/**
+ * Component which displays all data requirement dependencies for a specified resource
+ * and displays them as resource cards that contain all required information and link to their
+ * respective resources if a link exists
+ */
+function Dependencies(props: { relatedArtifact: fhir4.RelatedArtifact }) {
```
mgramigna, line 38: Really small code styling nitpick: https://gist.github.com/mgramigna/bdc933fae2150216681c28d1f0f7aac7#props

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -50,6 +53,10 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     window.addEventListener('resize', handleResize);
   }, []);
 
+  useEffect(() => {
+    setActiveTab('json');
+  }, [jsonData.id]);
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:58]: I'm a bit confused on this `useEffect`. We are subscribing to changes on the value of `jsonData.id`, which is obtained through serverside props, and then setting the active tab to JSON. Is there a need to do this at all? The default value of the tab is already `json`, and the id should never change client side since changing of `jsonData` would happen due to a full re-load of the page server side

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -80,6 +87,32 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+        if (displayA < displayB) {
+          return -1;
+        }
+        if (displayA > displayB) {
+          return 1;
+        }
+      }
+      return 0;
+    });
+  }, [dataRequirements?.relatedArtifact]);
```
Comment (mgramigna) [inline app/src/pages/[resourceType]/[id].tsx:114]: This works perfectly fine, but a slight nitpick is that I think it might be more intuitive if we don't sort the relatedArtifact within a useEffect whose dependency is itself the relatedArtifact. It could be a bit confusing if one is assuming that the dependencies are rendered as is and then having to trace down where the side effect of the `.sort` call actually happens on that original array.

My suggestion is to make use of the fact that `.sort` in javascript also [returns a reference to the sorted array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#return_value), and be more expressive about the variable names. So after the `dataRequirements` variable is initialiazed after the trpc useQuery is done, you could do something like:

```ts
  const sortedDependencies = dataRequirements?.relatedArtifact?.sort((a, b) => {
      // Your original code
  }) ?? []; // default to empty array until we have actual data to sort through
```

Then, when you go to render them, instead of

```tsx
                {dataRequirements.relatedArtifact.map((relatedArtifact, index) => (
                  <Dependency key={index} relatedArtifact={relatedArtifact} />
                ))}
```

You can now do

```tsx
                {sortedDependencies.map((relatedArtifact, index) => (
                  <Dependency key={index} relatedArtifact={relatedArtifact} />
                ))}
```

Now looking at the code, it is clear that the dependencies are expected to be iterated on in a sorted order, and I can easily take a look at the variable declaration of `sortedDependencies` to see how that is done.

app/src/components/DependencyCards.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  let dependencyInfo: DependencyInformation = {};
+
+  if (resourceLink?.includes('Library')) {
+    const resourceArr: string[] = resourceLink.substring(resourceLink.indexOf('Library')).split('|');
+    dependencyInfo = { type: 'Library', link: resourceArr[0] };
+  } else if (resourceLink?.includes('Measure')) {
+    const resourceArr: string[] = resourceLink.substring(resourceLink.indexOf('Measure')).split('|');
+    dependencyInfo = { type: 'Measure', link: resourceArr[0] };
+  }
```
mgramigna, line 51: If I understand what this strategy is doing, we are taking the text in brackets:

`http://somethingsomething.com/Library/[SOME TEXT]|0.0.0`

And using that text as the "`id`" that we route to to view the details of a resource, since the way our app is set up, the routing responds to `/[resourceType]/[id]`. Unfortunately, I don't think this strategy will work all of the time. 

Canonicals for FHIR resources can theoretically be anything. They _should_ be of the format that this code works on, but they might not always. I could have a measure whose URL is something like `http://example.com/Measure/MattRules123`, but the actual FHIR Measure resource that exists in the measure repository service has a completely different server-defined `.id` property that is not `MattRules123`.

A way to get around this might be out of scope for your task, so I might suggest we create another task to handle this, but I think we need to add logic to this app that can handle trying to view resource details using the canonical URL of the resource as the identifying information. So a URL like `localhost:3001/Measure?url=http://example.com/Measure/MattRules123` should be able to look up the measure in the service by canonical and then route to it.

OR we just have this be part of the existing search page instead or something. Either way, your code doesn't have to change approaches for now, but I would maybe suggest adding a comment above this block that talks about what I just mentioned and we can have a TODO to address it in later tasking

mgramigna (changes requested):

app/src/pages/[resourceType]/[id].tsx
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx:58]: I definitely believe there is a better way to do this, but the reason I included this` useEffect` is due to an error I was running into whenever a user attempted to navigate to a `dependency artifact` using its link. If the dependency artifact didn't have a dependencies tab of it's own because the "get data requirements" button wasn't clicked yet, the user would be brought to a blank page because it would be on a "dependencies" tab that didn't yet exist. It looked like this for example: 
Screenshot 2023-07-07 at 12 04 48 PM

So I tried to remedy this via the useEffect so whenever the user clicks a dependency artifact's link and gets routed to another page, it would automatically set the JSON tab as active.

app/src/components/DependencyCards.tsx
BriannaVH: Definitely a great point, I didn't even consider this when first starting on the task! I agree it might be a bit out of scope however, so I added the TODO like you suggested :)

lmd59 (commented):
I know the `Get Data Requirements` button was under consideration for removal. Did we decide to keep it? Should it be renamed? In the same vein, do we want to always show dependencies if they're available (so we don't have to hit the button again to show them after a reload)?
Screenshot 2023-07-11 at 11 54 38 AM

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -80,6 +98,50 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+        const displayFirst = firstElement.display?.toUpperCase();
+        const displaySecond = secondElement.display?.toUpperCase();
+        if (displayFirst && displaySecond) {
+          if (
+            firstElement.resource?.includes('Library') ||
+            firstElement.resource?.includes('Measure') ||
+            secondElement.resource?.includes('Library') ||
+            secondElement.resource?.includes('Measure')
+          ) {
+            return 1;
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx:116]: Good point! I think originally, the reason as to why I did this was because I was thinking it might be best to place all of the resources that have links at the very top so the user can easily select them. However, I didn't sort the resources with links alphabetically so I edited my current code to reflect that. I was curious though, do you think it would be better to sort everything solely alphabetically or to place the artifacts with links first and then sort everything else after? I have screenshots of both attached: 

Dependencies with links placed first:
Screenshot 2023-07-12 at 1 20 05 PM

Dependencies sorted regularly: 
Screenshot 2023-07-12 at 1 21 25 PM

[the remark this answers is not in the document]
Comment (lmd59) [reply app/src/pages/[resourceType]/[id].tsx:116]: I think that artifacts with links first is a good idea! Option A looks good to me :-)

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -80,6 +98,50 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
 
+  //Sorts dependencies alphabetically based on their display property. If the the
+  //dependency is a library/measure, however, it will always take precedence in sorting to ensure all
+  //dependencies with links are always shown first
+  const sortDependencies = useCallback(() => {
+    if (dataRequirements?.relatedArtifact) {
+      dataRequirements.relatedArtifact.sort((firstElement, secondElement) => {
+        const displayFirst = firstElement.display?.toUpperCase();
+        const displaySecond = secondElement.display?.toUpperCase();
+        if (displayFirst && displaySecond) {
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx:109]: Unfortunately the displayFirst and displaySecond variables are potentially undefined and so it will throw an error if we don't check they exist first

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -80,6 +98,50 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
     }
   );
 
+  //Sorts dependencies alphabetically based on their display property. If the the
+  //dependency is a library/measure, however, it will always take precedence in sorting to ensure all
+  //dependencies with links are always shown first
+  const sortDependencies = useCallback(() => {
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx:81]: Due to the useEffect on line 118-129, the search results will display all dependencies whenever the sortDependencies variable is called. So if the user searches something then it will display the results briefly and then revert back to showing all dependencies at once. The way I remedied this is by using the useCallBack to help make sure the search functionality actually works.

app/src/pages/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (lmd59) [reply app/src/pages/[resourceType]/[id].tsx:86]: Oh- I mean, what about the case where they're both undefined? Do we have any logic to handle that?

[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/[resourceType]/[id].tsx:88]: Ohhh I understand, my bad! I tried remedying this by sorting by the resource property if the display property doesn't exist for one of the elements. Do you think this is a decent solution?

mgramigna (approved):

PR #48, opened 2023-07-07
Add UI input for selecting a main logic library while authoring a mea…

lmd59:
…sure (and miscellaneous README updates)

# Summary
Adds dropdown in Measure authoring edit screen that allows selection of an authoring draft library as the `.library` main logic library for that measure

## New behavior
Measure authoring screen has library options dropdown, which can be used to update the Measure `.library` field.

## Code changes

- Changes to two `README`s that I noticed might be nice or necessary (mostly came up in reviewing other folks' PRs)
- `[id].tsx` adds a mantine `Select` component and `library` fields so that the Measure object can be mutated with the selected result. The `.library` field is an array of exactly one canonical string based on the library url, and the select default is `null` rather than an empty string, so there are a couple of differences in handling as compared with the other editable fields.

# Testing guidance

- `npm run check:all`
- `npm run start:all`
- Go to http://localhost:3001/
- Navigate to the authoring mode and create a new `Measure` draft resource (or navigate to the Authoring Measures list and choose to edit an existing draft measure)
- Select the `library` dropdown field. Available draft libraries should be shown in alphabetical order by `url` or shown at the end as disabled by `id` if a url is not available (id only libraries are shown so that the user knows they exist, but disabled because a url is required to select it). You can select a library or begin typing to filter the list to a desired library subset.
  - If there are no draft library options, the dropdown should indicate "No libraries available".
Screenshot 2023-07-07 at 2 55 34 PM

- Select `Submit` once the library is chosen (and any other desired text field updates are made). Submit should update successfully and show the corresponding library array in the json on the right.
Screenshot 2023-07-07 at 2 58 25 PM

- Try different combinations of selecting a library, selecting no library, changing other fields, etc.
- Draft Library editing should remain unaltered and continue to function

### DIFF
--- a/app/src/pages/authoring/[resourceType]/[id].tsx
+++ b/app/src/pages/authoring/[resourceType]/[id].tsx
@@ -1,10 +1,10 @@
 import { trpc } from '@/util/trpc';
-import { Button, Center, Divider, Grid, Paper, Select, Stack, Text } from '@mantine/core';
+import { Button, Center, Divider, Grid, Group, Paper, Select, Stack, Text, Tooltip } from '@mantine/core';
 import { useState, useEffect } from 'react';
 import { useRouter } from 'next/router';
 import { Prism } from '@mantine/prism';
 import { notifications } from '@mantine/notifications';
-import { AlertCircle, CircleCheck } from 'tabler-icons-react';
+import { AlertCircle, CircleCheck, InfoCircle } from 'tabler-icons-react';
 import { ArtifactResourceType } from '@/util/types/fhir';
 import ArtifactFieldInput from '@/components/ArtifactFieldInput';
 
@@ -206,7 +206,16 @@ export default function ResourceAuthoringPage() {
             <ArtifactFieldInput label="description" value={description} setField={setDescription} />
             {resourceType === 'Measure' && (
               <Select
-                label="library"
+                label={
+                  <Group spacing="xs">
+                    library
+                    <Tooltip label="Only draft libraries with a valid url may be selected.">
+                      <div>
+                        <InfoCircle size="1rem" style={{ display: 'block', opacity: 0.5 }} />
+                      </div>
+                    </Tooltip>
+                  </Group>
+                }
                 value={library}
                 onChange={setLibrary}
                 placeholder="Select main logic library"
--- a/app/src/pages/authoring/[resourceType]/[id].tsx
+++ b/app/src/pages/authoring/[resourceType]/[id].tsx
@@ -1,5 +1,5 @@
 import { trpc } from '@/util/trpc';
-import { Button, Center, Divider, Grid, Paper, Stack, Text } from '@mantine/core';
+import { Button, Center, Divider, Grid, Paper, Select, Stack, Text } from '@mantine/core';
 import { useState, useEffect } from 'react';
 import { useRouter } from 'next/router';
 import { Prism } from '@mantine/prism';
@@ -14,6 +14,7 @@ interface DraftArtifactUpdates {
   name?: string;
   title?: string;
   description?: string;
+  library?: string[] | null;
 }
 
 export default function ResourceAuthoringPage() {
@@ -26,6 +27,7 @@ export default function ResourceAuthoringPage() {
   const [name, setName] = useState('');
   const [title, setTitle] = useState('');
   const [description, setDescription] = useState('');
+  const [library, setLibrary] = useState<string | null>(null);
 
   const ctx = trpc.useContext();
 
@@ -56,6 +58,9 @@ export default function ResourceAuthoringPage() {
           savedIdentifierSystem = resource.identifier[0].system;
         }
       }
+    }
+    if (resource?.resourceType === 'Measure' && library !== resource?.library?.[0]) {
+      return true;
     }
     return (
       url !== (resource?.url ?? '') ||
@@ -98,6 +103,9 @@ export default function ResourceAuthoringPage() {
     if (resource?.description) {
       setDescription(resource.description);
     }
+    if (resource?.resourceType === 'Measure' && resource?.library?.[0]) {
+      setLibrary(resource.library?.[0]);
+    }
   }, [resource]);
 
   const resourceUpdate = trpc.draft.updateDraft.useMutation({
@@ -126,7 +134,8 @@ export default function ResourceAuthoringPage() {
     identifierSystem: string,
     name: string,
     title: string,
-    description: string
+    description: string,
+    library: string | null
   ) {
     const additions: DraftArtifactUpdates = {};
     const deletions: DraftArtifactUpdates = {};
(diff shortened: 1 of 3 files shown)


elsaperelli (commented):
I this is potentially out of scope for your task because I believe it was something that should have been including in previous draft artifacts tasks but nonetheless...in the spec [here](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#draft) says "t is an error to create multiple drafts of the same artifact by URL." It looks like we _do_ allow for this, because I am able to create multiple draft libraries with url "http://ecqi.healthit.gov/ecqms/Library/HospiceFHIR4":

image

This results in the following behavior in your task:

https://github.com/projecttacoma/measure-repository/pull/48

Not sure if we want to fix this in your task, or add a backlog fix for it, but wanted to point it out!

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -57,6 +59,9 @@ export default function ResourceAuthoringPage() {
         }
       }
     }
+    if (resource?.resourceType === 'Measure' && library !== resource?.library?.[0]) {
```
Comment (elsaperelli) [inline app/src/pages/authoring/[resourceType]/[id].tsx:62]: Thoughts on moving this into the the following return statement boolean? Or do you think it's more readable here?

Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:62]: Regardless of where it is, I do think the logic here has to be ever so slightly altered. Super specific situation, but if the user removes the library and clicks submit, then the submit button stays enabled. The submit button will stay enabled regardless of changes if the library is not defined (which I don't think is the behavior we want?).

Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:62]: I think it's because `library` is null and the library ON the resource is `undefined`...maybe changing the type of `library` would fix this?

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -144,10 +153,37 @@ export default function ResourceAuthoringPage() {
...
   }
 
+  // set up main library options
+  let libOptions: { value: string; label: string; disabled: boolean }[] = []; // default to empty
+  if (resourceType === 'Measure') {
+    const { data: libraries } = trpc.draft.getDrafts.useQuery('Library' as ArtifactResourceType);
+    if (libraries) {
+      libOptions = libraries?.map(l => {
+        if (l.url) {
+          // prioritizes use of url/version
```
Comment (elsaperelli) [inline app/src/pages/authoring/[resourceType]/[id].tsx:168]: Do we need to include `version` stuff here at all? Since we are only looking at draft libraries I don't think there will ever be versions.

elsaperelli (changes requested):
Just a few inline comments but other than that looks great!

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (lmd59) [reply app/src/pages/authoring/[resourceType]/[id].tsx:168]: I kind of left it in in case we were ever dealing with non-draft libraries or reusing it for anything that has version. This is technically how the canonical reference works: we just happen to know that there won't be a version in the current context. But can definitely remove if we don't think it's ever going to be useful!

BriannaVH (commented):
Overall looks great, just added some minor suggestions/edits for you!

 I ran into a similar issue as Elsa where, because we have multiple draft libraries with the same url, there are multiple select options with the same library url. I think that is probably a bit out of scope for you to tackle, but it's a good thing to take note of :) Maybe a temporary quick fix is to get rid of duplicates in libOptions, but that also doesn't take into account the fact that these duplicate draft libraries might have differing properties.

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -144,10 +153,37 @@ export default function ResourceAuthoringPage() {
...
+          const val = `${l.url}${l.version ? `|${l.version}` : ''}`;
+          return { value: val, label: val, disabled: false };
+        } else {
+          // uses id (assumed to exist in this context), but option disabled if no url exists (required for canonical reference)
+          const val = l.id ?? '';
+          return { value: val, label: val, disabled: true };
```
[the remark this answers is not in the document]
Comment (lmd59) [reply app/src/pages/authoring/[resourceType]/[id].tsx:174]: cool idea- let me see if I can make that happen!

lmd59:
> Overall looks great, just added some minor suggestions/edits for you!
> 
> I ran into a similar issue as Elsa where, because we have multiple draft libraries with the same url, there are multiple select options with the same library url. I think that is probably a bit out of scope for you to tackle, but it's a good thing to take note of :) Maybe a temporary quick fix is to get rid of duplicates in libOptions, but that also doesn't take into account the fact that these duplicate draft libraries might have differing properties.

I think we have a future task to address the duplicate URL issue!

lmd59:
Quick note: I'm seeing a bug that is not specific to this PR, where if we clear all optional fields, we get an error on submission (might look into this too if I have time)
Screenshot 2023-07-10 at 2 29 59 PM

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/authoring/[resourceType]/[id].tsx:166]: Gotcha! Doesn't hurt to leave in

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -64,7 +66,8 @@ export default function ResourceAuthoringPage() {
       (identifierValue.trim() !== '' && identifierSystem !== savedIdentifierSystem) ||
       name !== (resource?.name ?? '') ||
       title !== (resource?.title ?? '') ||
-      description !== (resource?.description ?? '')
+      description !== (resource?.description ?? '') ||
+      (resource?.resourceType === 'Measure' && (library ? library !== resource?.library?.[0] : resource?.library?.[0]))
```
Comment (elsaperelli) [inline app/src/pages/authoring/[resourceType]/[id].tsx:70]:
```suggestion
      (resource?.resourceType === 'Measure' && (library ? library !== resource.library?.[0] : resource.library?.[0]))
```
I think we know `resource` exists because of the first part of this line

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -98,6 +101,9 @@ export default function ResourceAuthoringPage() {
     if (resource?.description) {
       setDescription(resource.description);
     }
+    if (resource?.resourceType === 'Measure' && resource?.library?.[0]) {
```
Comment (elsaperelli) [inline app/src/pages/authoring/[resourceType]/[id].tsx:104]:
```suggestion
    if (resource?.resourceType === 'Measure' && resource.library?.[0]) {
```
Same here I think

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -144,10 +151,37 @@ export default function ResourceAuthoringPage() {
...
 
     return [additions, deletions];
   }
 
+  // set up main library options
+  let libOptions: { value: string; label: string; disabled: boolean }[] = []; // default to empty
+  if (resourceType === 'Measure') {
+    const { data: libraries } = trpc.draft.getDrafts.useQuery('Library' as ArtifactResourceType);
+    if (libraries) {
+      libOptions = libraries?.map(l => {
```
Comment (elsaperelli) [inline app/src/pages/authoring/[resourceType]/[id].tsx:164]:
```suggestion
      libOptions = libraries.map(l => {
```

elsaperelli (changes requested):

elsaperelli (approved):

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -170,6 +204,28 @@ export default function ResourceAuthoringPage() {
             <ArtifactFieldInput label="name" value={name} setField={setName} />
             <ArtifactFieldInput label="title" value={title} setField={setTitle} />
             <ArtifactFieldInput label="description" value={description} setField={setDescription} />
+            {resourceType === 'Measure' && (
+              <Select
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:208]: I did notice that if you select a library who has duplicate options, the duplicates get highlighted as being selected as well. I figure this is mainly a result of the fact that we can have duplicates in the first place, so I don't think it's something you need to fix. Just figured it's worth noting! 
Screenshot 2023-07-12 at 12 19 58 PM

app/src/pages/authoring/[resourceType]/[id].tsx
```diff
@@ -170,6 +204,28 @@ export default function ResourceAuthoringPage() {
...
+                label={
+                  <Group spacing="xs">
+                    library
+                    <Tooltip label="Only draft libraries with a valid url may be selected.">
```
Comment (BriannaVH) [inline app/src/pages/authoring/[resourceType]/[id].tsx:212]: I LOVE the tooltip! I think it's a good way to inform the user why certain options are disabled and it looks really good. Great work!

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (BriannaVH) [reply app/src/pages/authoring/[resourceType]/[id].tsx:208]: On the same note, I also noticed that this error message pops up whenever I click the select component for an artifact with duplicate libraries: 
Screenshot 2023-07-12 at 12 24 45 PM
While adding a key to your select component doesn't fix this error message because duplicate libraries would have the same key, I think it would be a good addition for whenever the duplicate problem is gone!

BriannaVH (approved):
Great job! I think this addition looks super great! The comments I did add are about minor problems which arose due to the duplicate artifacts issue, so it's mainly just something to be aware of rather than anything you have to fix currently! Love the tooltip addition!

app/src/pages/authoring/[resourceType]/[id].tsx
Comment (lmd59) [reply app/src/pages/authoring/[resourceType]/[id].tsx:208]: Duplicate urls should be addressed in ticket 1344 "Add URL constraint on drafting resources", which pertains to [spec](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#draft) "It is an error to create multiple drafts of the same artifact by URL."

PR #51, opened 2023-07-13
Artifact Review Page

BriannaVH:
# Summary
Users can now click a `report icon` to route themselves to a new review page for an artifact!

## New behavior
This PR allows a user to click the `report icon` on a resource card for an artifact to route themselves to a review page. In the future, this review page will be used  to add review comments and visualize review comments of an artifact. All repository/authoring artifacts should have a review page.  I also added a `review tab` that the user can select other than the  `repository/authoring` tabs that will route the user to a blank review page when selected.

## Code changes
`app/src/pages/review/[resourceType]/[id].tsx` - The components in this file are what the user sees when they click on the review icon of an artifact.
`Report` — I ended up using the `report icon` as the icon that the user must click in order to route themselves to the new page
`Tabs` — I added two tabs to the review page. In the future, I envision a user will toggle between the two tabs depending on if they want to add a review comment or visualize the existing comments
`Tooltip` —A tooltip pops up when the user hovers on the `report icon`
`Review tab` — I added a review tab that becomes highlighted when the user goes to review an artifact via clicking the icon. If the review tab is selected by itself, the user will just be brought to a blank page with no content currently.

# Testing guidance
`Npm run check:all
Npm run start:all`

Navigate to the repository portion of the website and ensure that every library/measure artifact has a `report icon` in their `resource card`. When the user hovers over it, the report icon tooltip should pop up. Upon clicking the report icon,  the url of the page the user is routed to should start with `/review/` and have `authoring = undefined` at the end. The `Review tab` should now be highlighted orange and the JSON of the artifact should be displayed in the review page. There should be two tabs on the page that, when the user toggles between them, still allows the JSON panel to be viewed. Now repeat the same process for an artifact in the `authoring` portion of the site. The functionality should be the same except for the fact that the url of the review page should have `authoring = true` at the end. When the user clicks on the review tab itself, they should see a coming soon page.

With regard to the navigation bar of the review page, it is essentially a combined version of the two navigation bars in the repository/authoring tabs. The user can navigate to either the repository/authoring part of the site based on whichever link they click.

Some questions for reviewers:
1. Do you like the review page format where the JSON is always displayed regardless of which tab is selected? 
2. Do you like the navigation bar for the review page? If not, do you have any suggestions? Is a nav bar necessary at all?


app/src/components/ResourceInfoCard.tsx
```diff
@@ -118,6 +118,18 @@ export default function ResourceInfoCard({ resourceInfo, authoring }: ResourceIn
...
               </Tooltip>
             </Link>
+            <Link
+              href={{
+                pathname: `/review/${resourceInfo.resourceType}/${resourceInfo.id}`,
+                query: { authoring: `${authoring}` }
+              }}
+            >
+              <Tooltip label={authoring ? 'Review Draft Artifact' : 'Review Artifact'} openDelay={1000}>
+                <ActionIcon radius="md" size="md" variant="subtle" color="gray">
```
elsaperelli, line 128: Maybe you can change the color of this so that it is not the same as the edit icon? They don't look super distinct at the moment:
image

Maybe one can be blue? Maybe that will look bad though, play around with what you think looks best!

app/tsconfig.json
```diff
@@ -12,6 +12,6 @@
       "@/*": ["./src/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "src/pages/review", "src/pages/review"],
```
elsaperelli, line 15:
```suggestion
  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
```
I don't think we need these anymore since we are no longer added a folder to the authoring folder, but correct me if I'm wrong.

app/src/pages/_app.tsx
```diff
@@ -62,19 +63,20 @@ function App({ Component, pageProps }: AppProps) {
           /** Consistent navbar shows available resources as regular content page changes to drill into details */
           navbar={
             <Navbar width={{ base: '320px' }}>
-              {router.pathname.startsWith('/authoring') ? (
+              {router.pathname.startsWith('/authoring') && (
                 <>
                   <Navbar.Section pt={18}>
                     <Center>
-                      <Text c="gray">Draft Artifacts</Text>
+                      <Text c="gray">Browse Measure Repository</Text>
```
elsaperelli, line 70:
```suggestion
                      <Text c="gray">Draft Artifacts</Text>
```
I think we want to keep this what it was before so that's it's clear they are draft artifacts and not artifacts in the measure repository.

app/src/pages/_app.tsx
```diff
@@ -89,12 +91,33 @@ function App({ Component, pageProps }: AppProps) {
...
+                        <Text c="gray">Browse Measure Repository</Text>
+                      </Center>
+                      <Divider my="md" />
+                      <Navbar.Section grow>
+                        <ServiceResourceButtons />
+                      </Navbar.Section>
+                    </Navbar.Section>
+                    <Navbar.Section grow pt={18}>
+                      <Center>
+                        <Text c="gray">Browse Authoring Repository</Text>
```
elsaperelli, line 113:
```suggestion
                        <Text c="gray">Draft Artifacts</Text>
```

app/src/pages/_app.tsx
```diff
@@ -89,12 +91,33 @@ function App({ Component, pageProps }: AppProps) {
...
                     <ServiceResourceButtons />
                   </Navbar.Section>
                 </>
               )}
+              {router.pathname.startsWith('/review') && (
+                <>
+                  <Stack>
+                    <Navbar.Section pt={18}>
+                      <Center>
+                        <Text c="gray">Browse Measure Repository</Text>
```
elsaperelli, line 104:
```suggestion
                        <Text c="gray">Measure Repository Artifacts</Text>
```

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,84 @@
...
+export default function CommentPage({ jsonData }: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  const jsonResourceType = jsonData.resourceType;
+  const router = useRouter();
+  const { resourceType: draftResourceType, id: draftID } = router.query;
+  const { authoring } = router.query;
+
+  const { data: resource } = trpc.draft.getDraftById.useQuery({
+    id: draftID as string,
+    resourceType: draftResourceType as ArtifactResourceType
+  });
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:22]: After chatting with Matt about this task, we decided that it would make more sense and would help us later on if we don't use `getServerSideProps` in this component and use `trpc` in order to fetch an artifact from the measure repository. This requires adding a procedure similar to this one in `service.ts` (this one is defined in `draft.ts`). Once you write a `getArtifactByid` procedure in `service.ts`, then you can use it here just like the above! So it will be something like the following:

```
if (authoring === true) {
     // we will want to fetch the artifact from the draft artifact db
     const { data: resource } = trpc.draft.getDraftById.useQuery...
} else {
    // authoring will be false, therefore we will want to fetch the artifact from the measure repository
    // we will want to do this with a trpc procedure that you will need to write in service.ts and then call it here
    const { data: resource } = trpc.service.getDraftById.useQuery...
}
```

Does that make sense? Happy to help with the `trpc procedures` definitions :)

app/src/components/ResourceInfoCard.tsx
```diff
@@ -39,6 +39,8 @@ export default function ResourceInfoCard({ resourceInfo, authoring }: ResourceIn
   const ctx = trpc.useContext();
   const [isConfirmationModalOpen, setIsConfirmationModalOpen] = useState(false);
 
+  authoring = authoring === undefined ? false : true;
+
```
elsaperelli, line 43: Rather than doing this here, if you look at where the `ResourceCards` component is called in `src/pages/authoring/[resourceType]/index.tsx`, you will see that it is called with `authoring={true}`. You can do this where `ResourceCards` is rendered in `src/pages/[resourceType]` and `src/pages/[resourceType]/search-result.tsx` with `authoring={false}`

app/src/components/ResourceInfoCard.tsx
```diff
@@ -106,19 +108,31 @@ export default function ResourceInfoCard({ resourceInfo, authoring }: ResourceIn
...
+              </Tooltip>
+            </Link>
+            <Link
+              href={{
+                pathname: `/review/${resourceInfo.resourceType}/${resourceInfo.id}`,
+                query: { authoring: `${authoring}` }
+              }}
+            >
+              <Tooltip label={authoring === true ? 'Review Draft Resource' : 'Review Resource'} openDelay={1000}>
+                <ActionIcon radius="md" size="md" variant="subtle" color="#339AF0">
```
elsaperelli, line 130:
```suggestion
                <ActionIcon radius="md" size="md" variant="subtle" color="blue">
```
I would just change this to blue- we try to use the Mantine presents whenever possible to make things simpler.

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,68 @@
...
+import { trpc } from '@/util/trpc';
+
+/**
+ * Component which renders a page that displays the JSON data of a resource. Also will eventually
+ *  provide the user with the ability to make review comments and visualize previous review comments.
+ */
+export default function CommentPage() {
+  const router = useRouter();
+  const { resourceType: resourceType, id: resourceID } = router.query;
+  const { authoring } = router.query;
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:14]: I think you can combine these two

app/src/server/trpc/routers/service.ts
```diff
@@ -54,6 +54,14 @@ export const serviceRouter = router({
       return resource as fhir4.Library;
     }),
 
+  getArtifactById: publicProcedure
+    .input(z.object({ resourceType: z.enum(['Measure', 'Library']), id: z.string() }))
+    .query(async ({ input }) => {
+      const res = await fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/${input.resourceType}/${input.id}`);
+      const resource = await res.json();
+      return resource as fhir4.Library;
```
elsaperelli, line 62: So I think we want to to return `resource` as `FhirArtifact` (which is a type that we defined that is `Measure | Library`. This is because we are getting an Artifact by id and it can either be a Measure or a Library so we don't want to return it as a library every time. @sarahmcdougall whenever you review this, can you check me on this? Or do we want to specify it is a `fhir4.Library` if `resourceType  = 'Library` and `fhir4.Measure` if `resourceType = 'Measure'` ?

elsaperelli (changes requested):
Great stuff! Just some small inline comments (you can wait to address the one I tagged Sarah in, I am curious to hear what she thinks)

app/src/components/ResourceInfoCard.tsx
sarahmcdougall: I agree with Elsa's approach - if we set `authoring` to true or false when passing it into `ResourceInfoCard`, then we can avoid having to reassign the value of `authoring`, which is preferred 👍

app/src/pages/_app.tsx
```diff
@@ -89,12 +91,33 @@ function App({ Component, pageProps }: AppProps) {
...
+                        <ServiceResourceButtons />
+                      </Navbar.Section>
+                    </Navbar.Section>
+                    <Navbar.Section grow pt={18}>
+                      <Center>
+                        <Text c="gray">Draft Artifacts</Text>
+                      </Center>
+                      <Divider my="md" />
+                    </Navbar.Section>
+                    <DraftResourceButtons />
```
sarahmcdougall, line 117: The `DraftResourceButtons` and `ServiceResourceButtons` components both call the `SidebarButtonList` component, which routes the resource buttons to certain paths. With the way things are set up right now, when I am on the "review" page and click on one of the resource buttons, I am routed to either the "repository" or "authoring" mode, and this gets reflected in the header. Not sure if this was intended, but I would expect that when you click on one of the resource buttons, you stay on the review mode. 

Wanted to see if this was discussed yet and whether it makes sense to make changes in this PR or in a future PR when the review page is further implemented. If we decide to make the change in this PR, I imagine that means adding some sort of conditional statement for what route prefix to use

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,68 @@
...
+  let resource;
+
+  if (authoring == 'true') {
```
Comment (sarahmcdougall) [inline app/src/pages/review/[resourceType]/[id].tsx:16]:
```suggestion
  if (authoring === 'true') {
```

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <div>
+      <Center>
+        <Text c="gray" fz="xl">
+          {authoring === 'true'
+            ? `Reviewing Draft ${resourceType}/${resourceID}`
+            : `Reviewing ${resourceType}/${resourceID}`}
+        </Text>
+      </Center>
+      <Divider my="md" style={{ marginTop: '14px' }} />
```
Comment (sarahmcdougall) [inline app/src/pages/review/[resourceType]/[id].tsx:37]: To make better use of the `Divider` props, we should use `mt` instead of `style`. I think this should achieve the same styling:
```suggestion
      <Divider my="md" mt={14} />
```

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Grid>
+        <Grid.Col span={6}>
+          <Tabs variant="outline" defaultValue="addComments">
+            <Tabs.List>
+              <Tabs.Tab value="addComments">Add comment</Tabs.Tab>
+              <Tabs.Tab value="visualizeComments"> Visualize Comments</Tabs.Tab>
```
Comment (sarahmcdougall) [inline app/src/pages/review/[resourceType]/[id].tsx:43]: Nitpick (and I expect that this might change in the future anyway) but I think for simplicity it might be preferable to say "View Comments" instead of "visualize"

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,68 @@
...
+            <Tabs.Panel value="addComments" pt="xs">
+              Components to add review comment goes here
+            </Tabs.Panel>
+            <Tabs.Panel value="visualizeComments" pt="xs">
+              Components to visualize comments go here
+            </Tabs.Panel>
+          </Tabs>
+        </Grid.Col>
+        <Grid.Col span={6}>
+          <Space></Space>
```
Comment (sarahmcdougall) [inline app/src/pages/review/[resourceType]/[id].tsx:54]:
```suggestion
          <Space/>
```

app/src/server/trpc/routers/service.ts
sarahmcdougall: I think just returning as a `FhirArtifact` is fine since we never pull off the `resourceType` or any other fields from the resource that we render. Although if the review comments are added to the resource types in different ways then we probably would want to be more specific about the resource type. But I imagine that review comments are added in the same way to both measures and libraries

app/src/components/ResourceInfoCard.tsx
BriannaVH: Great idea!

sarahmcdougall (changes requested):
Great work with this! Let me know if you have any questions on my suggestions or if you want to discuss anything further (and let me know if anything I mentioned is out of scope for this task, since I was out of office when this task was created)

app/src/pages/review/index.tsx
```diff
@@ -0,0 +1,14 @@
+import { Center, Text, Divider } from '@mantine/core';
+
+export default function ResourceAuthoringPage() {
```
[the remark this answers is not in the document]
BriannaVH: That's definitely a good idea! I ran into a bit of a conundrum where when I actually implemented this, it worked locally on my machine and all the checks passed. However when I pushed this up it resulted in an error and after some investigation, I'm still not quite sure why it didn't work.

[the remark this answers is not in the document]
BriannaVH: I've temporarily changed it back to it's previous name, but any help would be appreciated! Thanks! :)

app/src/pages/_app.tsx
BriannaVH: Definitely a great idea! I implemented it so that when a user clicks a side button in the side bar in the review mode, they will remain in the review mode and see a list that is identical to what they would see in the authoring/repository mode. However when viewing the resource list in the review mode, the user can still be routed to the repository/authoring mode if they click a specific button on the card resource (i.e The draft resource edit button still routes to the authoring mode). Curious to know your thoughts about how it looks/ if it makes sense those buttons would still route the user to other modes.

app/src/components/DraftResourceButtons.tsx
```diff
@@ -2,11 +2,15 @@ import { Text, Paper, Stack, Center, Loader } from '@mantine/core';
 import { trpc } from '@/util/trpc';
 import SidebarButtonList from './SidebarButtonList';
 
+interface MyComponentProps {
```
elsaperelli, line 5: I would name this to something more descriptive, maybe like `DraftResourceButtonsProps`? (@sarahmcdougall do we need this interface? Or should we just pass `review: boolean` into the function directly?

app/src/components/ServiceResourceButtons.tsx
```diff
@@ -2,11 +2,14 @@ import { Text, Paper, Stack, Center, Loader } from '@mantine/core';
 import { trpc } from '@/util/trpc';
 import SidebarButtonList from './SidebarButtonList';
 
+interface MyComponentProps {
```
elsaperelli, line 5: Same comment as above, I would name this something else.

app/src/pages/[resourceType]/search-result.tsx
```diff
@@ -26,7 +26,7 @@ export default function ResourceSearchResultsPage({
       <Divider my="md" />
       {resourceInfo ? (
         <Stack align="center" pt={18}>
-          <ResourceCards resourceInfo={resourceInfo} resourceType={resourceType} />
+          <ResourceCards resourceInfo={resourceInfo} resourceType={'Measure'} authoring={false} />
```
Comment (elsaperelli) [inline app/src/pages/[resourceType]/search-result.tsx:29]:
```suggestion
          <ResourceCards resourceInfo={resourceInfo} resourceType={resourceType} authoring={false} />
```

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,67 @@
...
+            : `Reviewing ${resourceType}/${resourceID}`}
+        </Text>
+      </Center>
+      <Divider my="md" mt={14} />
+      <Grid>
+        <Grid.Col span={6}>
+          <Tabs variant="outline" defaultValue="addComments">
+            <Tabs.List>
+              <Tabs.Tab value="addComments">Add comment</Tabs.Tab>
+              <Tabs.Tab value="viewComments"> View Comments</Tabs.Tab>
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:43]:
```suggestion
              <Tabs.Tab value="viewComments">View Comments</Tabs.Tab>
```

elsaperelli (changes requested):
I am wondering what the best way to do routes to the `review` tab and associated pages are...so stayed turned for some comments about that. For now, I just have some small inline comments :)

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -0,0 +1,67 @@
...
+          <Space />
+          <Text c="gray" fz="sm">
+            Current JSON Content
+          </Text>
+          <Paper withBorder>
+            <Prism language="json" colorScheme="light" styles={{ scrollArea: { height: 'calc(100vh - 150px)' } }}>
+              {resource.data ? JSON.stringify(resource.data, null, 2) : ''}
+            </Prism>
+          </Paper>
+        </Grid.Col>
```
Comment (sarahmcdougall) [inline app/src/pages/review/[resourceType]/[id].tsx:63]: This doesn't require a change in this PR, but once we have the UI in place for adding comments, it might be nice to extend the width of the JSON content so that the user does not need to scroll as much when viewing the JSON. But this will depend on how the comments look once that functionality is in place :) just something to keep in mind for now

sarahmcdougall (approved):
Great work with this task! It's awesome to see this feature come to life! I like the new simplicity with the routing

elsaperelli (approved):
Great job! Thanks for being patient and flexible as we worked through what works best for routing!

PR #53, opened 2023-07-24
Add Artifact comments to a draft resource

BriannaVH:
# Summary
This PR allows users to add an artifact comment to a **draft** resource.

## New behavior

On the review page for an artifact, users will be greeted with a `use-form mantine` component in the `add comment` tab panel. Via the form, users can enter the following information: The comment, their name, [the comment type](https://build.fhir.org/ig/HL7/cqf-measures/CodeSystem-artifact-comment-type.html), and whether they would like to include the date or not. The exact current date/time is what will be added to the artifact comment if the user chooses to include the date. Amongst these 4 parameters, only the comment type and the actual comment are needed in order to submit the form. Upon submitting the form, the submit button should show a `loader icon` for 1 millisecond, the form properties should clear themselves, and the new comment should now be loaded in the JSON content. In the JSON itself, the new artifact comment should appear as a `fhir4.Extension array of objects` in another `fhir4.Extension array`. 

When the user enters input into all of the components in the form, the new extension array of objects should look like the following:
Screenshot 2023-07-24 at 5 27 32 PM
If the user opted to not include a name then the object with "user" as the url should not appear. If the user doesn't include the date then the object with the url equal to "authoredOn" should not appear.
Users can enter as many comments as they would like and the artifact comments will persist even if the user navigates away from the page. The user will be able to submit these forms in both authoring/repository mode, however the JSON content will only be permanently changed for draft artifacts.

## Code changes
`"@mantine/form"` --> I installed the "@mantine/form" dependency in order to add a form on this page.
`form` --> The form object is composed of all of the mantine components the user must enter input into in order to submit the form. It will double check that the user entered their input in the proper format.
`HoverCard` --> When the user hovers their mouse over the `IconInfoHexagonFilled` icon/ information icon next to the select mantine component, a hover card should pop up explaining the different comment types in more detail.
`updateDraft` --> In order to update a draft resource, I call the trpc procedure updateDraft. This is the reason why only draft resources have their JSON permanently changed.

# Testing guidance
```
Npm run check:all 
Npm run start:all
```
Navigate to the review page for a draft resource artifact. An extension array should already be present in the JSON, although it may have no content entered into it. First fill out the form with all of the information included. Once you click submit, the extension array in the JSON should have a new `fhir4.Extension array of objects` that looks like the above image and a success notification should pop up. Now, try entering as many other comments as you'd like with any combination of input fields. As long as the comment type and comment are filled out, the comment should have no errors when the user clicks submit. Otherwise, the mantine component  that caused the error will highlight red and some text will pop up about what went wrong. Lastly, navigate away from the draft artifact and return to the same review page, all of the comments should still appear in the JSON.

### DIFF
--- a/app/src/pages/review/[resourceType]/[id].tsx
+++ b/app/src/pages/review/[resourceType]/[id].tsx
@@ -26,7 +26,7 @@ import { useRouter } from 'next/router';
 import { trpc } from '@/util/trpc';
 import { IconStar, IconCalendar, IconInfoHexagonFilled } from '@tabler/icons-react';
 import { AlertCircle, CircleCheck, InfoCircle } from 'tabler-icons-react';
-import { hasLength, isNotEmpty, useForm } from '@mantine/form';
+import { isNotEmpty, useForm } from '@mantine/form';
 import { notifications } from '@mantine/notifications';
 
 interface DraftArtifactUpdates {
@@ -50,25 +50,30 @@ export default function CommentPage() {
   //Validates that all of the necessary form properties were entered by the user
   const form = useForm({
     initialValues: {
-      name: '',
       type: '',
-      comment: ''
+      comment: '',
+      name: ''
     },
+    //An error will be thrown if these fields aren't entered properly
     validate: {
-      name: hasLength({ min: 2, max: 10 }, 'Name must be 2-10 characters long'),
       type: isNotEmpty('Select the type of comment'),
       comment: isNotEmpty('Enter artifact comment')
     }
   });
 
+  //Currently we can only update draft artifact resources.
   const resourceUpdate = trpc.draft.updateDraft.useMutation({
     onSuccess: () => {
-      notifications.show({
-        title: 'Comment Successfully added!',
-        message: `${resourceType} Successful comment`,
-        icon: <CircleCheck />,
-        color: 'green'
-      });
+      //This if statement prevents the success notifaction from popping up in the unique case
+      //that the draft artifact has just had an extension array added to the JSON.
+      if (JSON.stringify(resource?.extension) !== undefined) {
+        notifications.show({
+          title: 'Comment Successfully added!',
+          message: `Comment Successfully added to ${resourceType}/${resourceID}`,
+          icon: <CircleCheck />,
+          color: 'green'
+        });
+      }
       ctx.draft.getDraftById.invalidate();
     },
     onError: e => {
@@ -83,7 +88,6 @@ export default function CommentPage() {
 
   function getResource() {
     if (authoring === 'true') {
-      //test here
       return trpc.draft.getDraftById.useQuery({
         id: resourceID as string,
         resourceType: resourceType as ArtifactResourceType
@@ -99,31 +103,34 @@ export default function CommentPage() {
   function parseUpdate(comment: string, type: string, userName: string, dateSelected: boolean) {
     const additions: DraftArtifactUpdates = {};
     const deletions: DraftArtifactUpdates = {};
-    let newExtensionObject: fhir4.Extension[] = [];
 
-    newExtensionObject.push({ url: 'type', valueCode: type });
-    newExtensionObject.push({ url: 'text', valueMarkdown: comment });
-    newExtensionObject.push({ url: 'user', valueString: userName });
+    const newExtensionObject: fhir4.Extension[] = [];
+    newExtensionObject.push({ url: 'type', valueCode: type }, { url: 'text', valueMarkdown: comment });
+
+    if (userName !== '') {
+      newExtensionObject.push({ url: 'user', valueString: userName });
+    }
     if (dateSelected === true) {
       const now = new Date();
       const isoString = now.toISOString();
       newExtensionObject.push({ url: 'authoredOn', valueDateTime: isoString });
     }
-    if (comment.trim() !== '') {
-      if (resource?.extension) {
-        resource.extension.push({
-          extension: newExtensionObject,
-          url: 'http://hl7.org/fhir/us/cqfmeasures/CodeSystem/artifact-comment-type'
-        });
-        additions['extension'] = resource.extension;
-      }
+    if (resource?.extension) {
+      resource.extension.push({
+        extension: newExtensionObject,
+        url: 'http://hl7.org/fhir/us/cqfmeasures/CodeSystem/artifact-comment-type'
+      });
+      additions['extension'] = resource.extension;
     }
     return [additions, deletions];
   }
 
   // useEffect to check if the artifact has an extension and adds it if it doesn't
+  /* eslint-disable react-hooks/exhaustive-deps */
   useEffect(() => {
+    console.log('in here');
     if (!resource?.extension) {
+      console.log('no extension');
       const additions: DraftArtifactUpdates = {};
       const deletions: DraftArtifactUpdates = {};
       additions['extension'] = [];
@@ -135,6 +142,7 @@ export default function CommentPage() {
       });
     }
   }, [resource]);
+  /* eslint-enable react-hooks/exhaustive-deps */
 
   return (
     <div>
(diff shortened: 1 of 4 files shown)


app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -1,30 +1,147 @@
...
+      resourceUpdate.mutate({
+        resourceType: resourceType as ArtifactResourceType,
+        id: resourceID as string,
+        additions: additions,
+        deletions: deletions
+      });
+    }
+  }, [resource]);
+  /* eslint-enable react-hooks/exhaustive-deps */
+
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:144]: We talked in person about this, but just putting this here for documentation purposes. I think we can remove the `useEffect` and do the check for is `resource.extension` exists in the `parseUpdate` function. I believe that this will not only be less code, but will also solve the weirdness we are seeing with `extension` being set to empty array.

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -34,16 +151,149 @@ export default function CommentPage() {
...
-              Components to add review comment goes here
+              <Box
+                component="form"
+                maw={1200}
+                mx="auto"
+                //This console.log is necessary because the onSubmit function has to use the '() => {}' format
+                //and it will cause an error if it is left empty.
+                onSubmit={form.onSubmit(values => {
+                  console.log(values);
+                })}
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:171]: This is so weird of the Mantine docs!!! But I fear we need it

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -34,16 +151,149 @@ export default function CommentPage() {
...
+                </Group>
+                <Textarea
+                  radius="lg"
+                  mt="md"
+                  minRows={15}
+                  maxRows={15}
+                  placeholder="Your Artifact comment"
+                  label="Artifact Comment"
+                  description="Add a comment to the resource"
+                  withAsterisk
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/review/[resourceType]/[id].tsx:232]: It's a bit weird, but because I have the info icon beside the select, I can't include the withAsterisk property without the UI looking very...odd. I made it so that there is an asterisk, but  I did this by adding a red text mantine component that contains the *

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -34,16 +140,149 @@ export default function CommentPage() {
...
+                          form.values.type,
+                          form.values.name,
+                          dateSelected
+                        );
+                        resourceUpdate.mutate({
+                          resourceType: resourceType as ArtifactResourceType,
+                          id: resourceID as string,
+                          additions: additions,
+                          deletions: deletions
+                        });
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:278]: I would put these in an `if` statement or otherwise something that ensures that they are only called when `authoring === true`. Right now we only want to support adding the comment on the JSON for draft artifacts, but by calling `parseUpdate` here, the `resource.extension` is being temporarily set in the JSON but not saved in it (which may be confusing for the user). Additionally, the `mutate` procedure that gets called which we actually don't want. It still gets called and is successful because it tries to add additional information to a collection in the draft database but with an `id` that doesn't exist (since it's getting the id from the artifact in the central measure repository). `collection.insertOne()` is still successful, but it doesn't add anything so the pop up is misleading. Let me know if you want to talk through this more!

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -1,28 +1,134 @@
...
+      type: isNotEmpty('Select the type of comment'),
+      comment: isNotEmpty('Enter artifact comment')
+    }
+  });
+
+  // Currently we can only update draft artifact resources.
+  const resourceUpdate = trpc.draft.updateDraft.useMutation({
+    onSuccess: () => {
+      // This if statement prevents the success notifaction from popping up in the unique case
+      // that the draft artifact has just had an extension array added to the JSON.
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:68]:
```suggestion
```

I think what I talked about in a later comment will prevent this from happening! So you can just delete this comment.

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -1,28 +1,134 @@
...
+    const newExtensionObject: fhir4.Extension[] = [];
+    newExtensionObject.push({ url: 'type', valueCode: type }, { url: 'text', valueMarkdown: comment });
+
+    if (userName !== '') {
+      newExtensionObject.push({ url: 'user', valueString: userName });
+    }
+    if (dateSelected === true) {
+      const now = new Date();
+      const isoString = now.toISOString();
+      newExtensionObject.push({ url: 'authoredOn', valueDateTime: isoString });
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:114]:
```suggestion
      const authoredDate = new Date();
      newExtensionObject.push({ url: 'authoredOn', valueDateTime: authoredDate.toISOString() });
```
I think this is a little cleaner/more descriptive!

elsaperelli (changes requested):

elsaperelli (approved):
Awesome job! Thanks for being flexible and only implementing this for the draft artifacts for now. Perhaps soon we will add this functionality for the artifacts on the measure repository. Looks great!

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -34,16 +137,151 @@ export default function CommentPage() {
...
+                    ref={ref}
+                    id="checkbox"
+                    color="white"
+                    mt="md"
+                    label="Include Date in Comment"
+                    {...form.getInputProps('date')}
+                    onChange={() => {
+                      setDateSelected(!dateSelected);
+                    }}
+                  />
```
Comment (sarahmcdougall) [inline app/src/pages/review/[resourceType]/[id].tsx:245]: Not something that necessarily needs to be changed right now, but I'm having a hard time thinking of a use case where the reviewer would choose not to have the date recorded. Maybe we should bring this up in slack to get some input from Hoss/Paul/others who are more familiar with the measure development process

sarahmcdougall (changes requested):
I have a few suggested changes and some small comments. Great work on this! I am super impressed with how the UI looks!!!

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -1,28 +1,131 @@
...
-    });
-  } else {
-    resource = trpc.service.getArtifactById.useQuery({
-      id: resourceID as string,
-      resourceType: resourceType as ArtifactResourceType
-    });
+    if (resource?.extension) {
+      resource.extension.push({
+        extension: newExtensionObject,
+        url: 'http://hl7.org/fhir/us/cqfmeasures/CodeSystem/artifact-comment-type'
```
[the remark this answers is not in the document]
Comment (elsaperelli) [reply app/src/pages/review/[resourceType]/[id].tsx:117]: Oh yes! That's right. It's like that in the spec example [here](https://build.fhir.org/ig/HL7/cqf-measures/Measure-EXMLogic-FHIR.json.html) too. Good catch

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -34,16 +137,151 @@ export default function CommentPage() {
...
+              >
+                <Group>
+                  <Select
+                    id="commentType"
+                    mt="md"
+                    clearable
+                    radius="lg"
+                    label={
+                      <Group spacing="xs">
+                        <Text>Comment Type</Text> <Text color="red">*</Text>
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/review/[resourceType]/[id].tsx:167]: Hmmm that is definitely a good point. Unfortunately, the only way that I could get the asterisk to replicate what the artifact comment component below looks like is via wrapping the * in a red text component. Would you have any recommendations on how to modify the withAsterisk component so it looks the same but still keep the information icon?

app/src/pages/review/[resourceType]/[id].tsx
Comment (BriannaVH) [reply app/src/pages/review/[resourceType]/[id].tsx:245]: Great point! Just made a thread in slack regarding this!

app/src/pages/review/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/review/[resourceType]/[id].tsx]: I played around with it a little bit and couldn't come up with a good solution... I think for now the code is fine as-is

sarahmcdougall (approved):
Thanks for making those changes! If anyone else comments on the Slack discussion regarding the date inclusion then we can make changes as appropriate. Otherwise, should be good to merge and we can make changes in the future once we have more input on use cases for adding artifact comments 👍

elsaperelli (approved):
Great job with this!

PR #55, opened 2023-07-31
View previous artifact comments on a draft resource!

BriannaVH:
# Summary
Users now will be able to view all previous review comments that were made on a **draft** resource in a human-readable format.

## New behavior
Now, when users navigate to a **draft** resource's review page, the "View Comments" panel will show the user every comment that was made on an artifact. The user will see the comments in chronological order, with the most recent comment appearing at the top of the page. When the user hovers their mouse over the date of a comment it will highlight blue and upon being clicked, will display the date in a different format. A comment may optionally not have a date or a name and in those instances, the comment will visually look different. 

## Code changes
`app/src/components/ArtifactTimeline.tsx `--> This creates a `timeline` of comments. Each `ArtifactComment` is a `timeline item`

`app/src/components/ArtifactComments.ts`x --> This formats the way a comment should appear
`Avatar` --> The comments will display an avatar icon. If the user included their name in the comment, the avatar icon will display the first letter of the user's name. Otherwise, it will show the default user icon instead.

# Testing guidance
```
npm run check:all
npm run start:all

```
Navigate to the review page of a draft artifact with no comments. The "View comments" panel should display red text that informs the user there are no comments. Now add a comment to the resource with all of it's properties included. It should appear in the view comments panel now, replacing the red text. The comment type should appear as the header and the first letter of the user's name should appear as the avatar's icon. Upon hovering one's mouse over the date of a comment, it should change to blue and when the user clicks on the date it will switch to a different format. Now continue adding comments to the resource and try different combinations of comment properties to see how the comment appears.  If the user doesn't include a name, their avatar icon should display a user icon rather than the first letter of their name. If the user doesn't include a date it simply won't appear. The most recent comment should always appear at the top.

### DIFF
--- a/app/src/components/ArtifactComments.tsx
+++ b/app/src/components/ArtifactComments.tsx
@@ -1,4 +1,5 @@
 import { Avatar, createStyles, Group, Paper, rem, Text, TypographyStylesProvider } from '@mantine/core';
+import parse from 'html-react-parser';
 import { useHover } from '@mantine/hooks';
 import { useEffect, useState } from 'react';
 
@@ -43,7 +44,7 @@ export default function ArtifactComments({ date, body, author }: ArtifactComment
   function getDate() {
     if (date) {
       const stringDate = new Date(date);
-      return `${stringDate.getMonth()}/${stringDate.getDate()}/${stringDate.getFullYear()}`;
+      return `${stringDate.getMonth() + 1}/${stringDate.getDate()}/${stringDate.getFullYear()}`;
     }
   }
 
@@ -86,7 +87,8 @@ export default function ArtifactComments({ date, body, author }: ArtifactComment
         </Group>
         <div style={{ wordBreak: 'break-word' }}>
           <TypographyStylesProvider className={classes.body}>
-            {body && <div className={classes.content} dangerouslySetInnerHTML={{ __html: body }} />}
+            {body && <div className={classes.content} />}
+            {body && parse(body)}
           </TypographyStylesProvider>
         </div>
       </div>
--- a/app/src/components/ArtifactComments.tsx
+++ b/app/src/components/ArtifactComments.tsx
@@ -87,8 +87,7 @@ export default function ArtifactComments({ date, body, author }: ArtifactComment
         </Group>
         <div style={{ wordBreak: 'break-word' }}>
           <TypographyStylesProvider className={classes.body}>
-            {body && <div className={classes.content} />}
-            {body && parse(body)}
+            {body && <div className={classes.content} /> && parse(body)}
           </TypographyStylesProvider>
         </div>
       </div>
--- a/app/src/components/ArtifactComments.tsx
+++ b/app/src/components/ArtifactComments.tsx
@@ -1,0 +1,95 @@
+import { Avatar, createStyles, Group, Paper, rem, Text, TypographyStylesProvider } from '@mantine/core';
+import { useHover } from '@mantine/hooks';
+import { useEffect, useState } from 'react';
+
+const useStyles = createStyles(theme => ({
+  card: {
+    borderRadius: 12,
+    border: `${rem(2)} solid ${theme.colors.gray[3]}`,
+    width: '800px'
+  },
+  comment: {
+    padding: `${theme.spacing.lg} ${theme.spacing.xl}`
+  },
+
+  body: {
+    borderRadius: 12,
+    paddingLeft: rem(54),
+    paddingTop: theme.spacing.sm,
+    fontSize: theme.fontSizes.sm
+  },
+
+  content: {
+    '& > p:last-child': {
+      marginBottom: 0
+    },
+    width: '800px'
+  }
+}));
+
+interface CommentHtmlProps {
+  date: string | undefined;
+  body: string | undefined;
+  author: string | undefined;
+}
+
+export default function ArtifactComments({ date: date, body, author }: CommentHtmlProps) {
+  const { classes } = useStyles();
+  const { hovered, ref } = useHover();
+  const [color, setColor] = useState('grey');
+  const [displayedDate, setDisplayedDate] = useState<string | undefined>(date);
+
+  //This changes the UTC date format into the 'mm/dd/yyyy' format and returns that
+  function getDate() {
+    if (date) {
+      const stringDate = new Date(date);
+      return `${stringDate.getMonth()}/${stringDate.getDate()}/${stringDate.getFullYear()}`;
+    }
+  }
+
+  const formattedDate = getDate();
+
+  function changeDate() {
+    if (displayedDate === date) {
+      setDisplayedDate(formattedDate);
+    } else {
+      setDisplayedDate(date);
+    }
+  }
+
+  //Changes the color of the date text depending on if the user is hovering over the text or not
+  useEffect(() => {
+    if (hovered === true) {
+      setColor('blue');
+    } else {
+      setColor('grey');
+    }
+  }, [hovered]);
+
+  return (
+    <Paper withBorder radius="md" className={classes.card} shadow="md">
+      <div className={classes.comment}>
+        <Group>
+          <Avatar src="h" alt="n" radius="xl" color="blue">
+            {author ? author.charAt(0).toUpperCase() : ''}
+          </Avatar>
+          <div style={{ wordBreak: 'break-word' }}>
+            <Text weight={500} fz="sm">
+              {author ? author : 'Guest'}
+            </Text>
+            <div ref={ref} style={{ cursor: 'pointer' }}>
+              <Text variant="subtle" fz="xs" onClick={changeDate} opacity={0.8} color={color}>
+                {date ? displayedDate : ''}
+              </Text>
+            </div>
+          </div>
+        </Group>
+        <div style={{ wordBreak: 'break-word' }}>
+          <TypographyStylesProvider className={classes.body}>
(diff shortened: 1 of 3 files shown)


app/src/components/ArtifactComments.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+}));
+
+interface CommentHtmlProps {
+  date: string | undefined;
+  body: string | undefined;
+  author: string | undefined;
+}
+
+export default function ArtifactComments({ date: date, body, author }: CommentHtmlProps) {
```
elsaperelli, line 36:
```suggestion
export default function ArtifactComments({ date, body, author }: CommentHtmlProps) {
```
I think

app/src/components/ArtifactComments.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { classes } = useStyles();
+  const { hovered, ref } = useHover();
+  const [color, setColor] = useState('grey');
+  const [displayedDate, setDisplayedDate] = useState<string | undefined>(date);
+
+  //This changes the UTC date format into the 'mm/dd/yyyy' format and returns that
```
elsaperelli, line 42:
```suggestion
  // This changes the UTC date format into the 'mm/dd/yyyy' format
```

app/src/components/ArtifactComments.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  function changeDate() {
+    if (displayedDate === date) {
+      setDisplayedDate(formattedDate);
+    } else {
+      setDisplayedDate(date);
+    }
+  }
+
+  //Changes the color of the date text depending on if the user is hovering over the text or not
```
elsaperelli, line 60:
```suggestion
  // Changes the color of the date text depending on if the user is hovering over the text or not
```

app/src/components/ArtifactComments.tsx
```diff
@@ -0,0 +1,95 @@
...
+    },
+    width: '800px'
+  }
+}));
+
+interface CommentHtmlProps {
+  date: string | undefined;
+  body: string | undefined;
+  author: string | undefined;
+}
```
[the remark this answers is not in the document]
elsaperelli: I would also maybe change this interface's name to be `ArtifactCommentsProps` so that it matches the `ArtifactComments` function. This means we will have to change the interface name in the following file, but I think this makes a little more sense.

app/src/components/ArtifactTimeline.tsx
```diff
@@ -0,0 +1,112 @@
+import { Center, ScrollArea, Space, Text, Timeline } from '@mantine/core';
+import { Message } from 'tabler-icons-react';
+import { useEffect, useState } from 'react';
+import ArtifactComments from './ArtifactComments';
+
+interface ArtifactCommentProps {
+  date: string | undefined;
+  body: string | undefined;
+  author: string | undefined;
+}
```
elsaperelli, line 10: I would just import the interface from `ArtifactComments.tsx` since they are the same.

app/src/components/ArtifactTimeline.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+interface ExtensionArray {
+  extensions: undefined | fhir4.Extension[];
+}
```
elsaperelli, line 14:
```suggestion
interface ArtifactTimelineProps {
  extensions?: fhir4.Extension[];
}
```
I think this name makes a little more sense and you can do the `?` in place of `undefined` again!

app/src/components/ArtifactTimeline.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+interface commentProps {
+  author: string | undefined;
+  date: string | undefined;
+  body: string | undefined;
+  type: string | undefined;
+}
```
elsaperelli, line 21: Looks like this is being used as a type, so maybe it would be better to define it as such?

app/src/components/ArtifactTimeline.tsx
```diff
@@ -0,0 +1,112 @@
...
+      setWindowHeight(window.innerHeight);
+    };
+    handleResize();
+    window.addEventListener('resize', handleResize);
+    return window.removeEventListener('resize', handleResize);
+  }, []);
+
+  if (extensionArr?.extensions) {
+    if (extensionArr.extensions.length === 0) {
+      return <div>{noCommentsAvailable()}</div>;
```
elsaperelli, line 57: Honestly, I don't think you need a function for `noCommentsAvailable()`. I would just return the contents of the function here.

BriannaVH: I definitely think it wouldn't be necessary to use a function if it was just one instance of the call but since I use it three times I think it helps to get rid of repetitive code :)

app/src/components/ArtifactComments.tsx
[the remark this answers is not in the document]
BriannaVH: Great catch ! Interfaces tend to confuse me a bit so this was very helpful :)

app/src/components/ArtifactTimeline.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <Center>
+        <Text color="red">
+          <i>No Comments Found for Artifact</i>{' '}
+        </Text>
+      </Center>
+    </div>
+  );
+}
+
+export default function ArtifactTimeline(extensionArr: ArtifactTimelineProps) {
```
elsaperelli, line 38:
```suggestion
export default function ArtifactTimeline({ extensions }: ArtifactTimelineProps) {
```
You can just get `extensions` directory so that you don't have to do `extensionArr?.extensions` below!

app/src/components/ArtifactTimeline.tsx
```diff
@@ -0,0 +1,102 @@
...
+    };
+    handleResize();
+    window.addEventListener('resize', handleResize);
+    return window.removeEventListener('resize', handleResize);
+  }, []);
+
+  if (extensionArr?.extensions) {
+    if (extensionArr.extensions.length === 0) {
+      return <div>{noCommentsAvailable()}</div>;
+    } else {
```
elsaperelli, line 53: I don't think we need this if statement. `extensionArr.extensions.forEach` won't add anything to `commentArray` if `extensionArry.extensions.length === 0`, so it will reach the `if (commentArray.length == 0)` statement and return the `noCommentsAvailable` message then.

elsaperelli (dismissed):

sarahmcdougall (changes requested):
I am blown away with how great the UI looks!

I have a few suggested changes - let me know if you have any questions!

app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -282,24 +287,21 @@ export default function CommentPage() {
...
+            </Grid.Col>
+            <Grid.Col span={6}>
+              <Space />
+              <Text c="gray" fz="sm">
+                Current JSON Content
+              </Text>
+              <Paper withBorder>
+                <Prism language="json" colorScheme="light" styles={{ scrollArea: { height: 'calc(100vh - 150px)' } }}>
+                  {resource ? JSON.stringify(resource, null, 2) : ''}
+                </Prism>
```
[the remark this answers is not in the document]
Comment (BriannaVH) [reply app/src/pages/review/[resourceType]/[id].tsx:299]: Hmmm I was also torn on this before I ended up getting rid of the prism content on the view comments panel. I do agree that the target property should show the code, but I was thinking we could do something vaguely similar to what git has in place when someone makes a comment on a specific part of the code? We could put the code that was referenced in a card and attach it somehow under the comment itself! So could look like the very comment you left! 
Screenshot 2023-08-01 at 1 49 57 PM

app/src/components/ArtifactTimeline.tsx
BriannaVH: Good point! Ill get rid of that unnecessary check. I still think the `noCommentsAvailable` function should stay, however, cause there are two instances where there could be no comments. One where the artifact doesn't have an extension array in the first place and one in which it does have an extension array but it doesn't have any actual comments.

app/src/components/ArtifactTimeline.tsx
BriannaVH: Great catch!

app/src/pages/review/[resourceType]/[id].tsx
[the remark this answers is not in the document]
Comment (sarahmcdougall) [reply app/src/pages/review/[resourceType]/[id].tsx:299]: I really like that idea!!! That could be a great follow-up task for when we implement the `target` field

app/src/components/ArtifactComments.tsx
```diff
@@ -0,0 +1,97 @@
...
+              <Text variant="subtle" fz="xs" onClick={changeDate} opacity={0.8} color={color}>
+                {date ? displayedDate : ''}
+              </Text>
+            </div>
+          </div>
+        </Group>
+        <div style={{ wordBreak: 'break-word' }}>
+          <TypographyStylesProvider className={classes.body}>
+            {body && <div className={classes.content} />}
+            {body && parse(body)}
```
sarahmcdougall, line 91:
```suggestion
            {body && <div className={classes.content}> {parse(body)} </div>}
```
Seems like we can combine these two lines since they both check for the same condition

sarahmcdougall (changes requested):
Just one more comment but otherwise looks good to me!

sarahmcdougall (approved):

PR #56, opened 2023-08-10
Release a draft artifact to the repository!

BriannaVH:
# Summary
Users can now release a draft artifact to the MRS!

## New behavior
Users can navigate to the “edit” page of a draft artifact and will see that there is now a release button on the bottom of the page. When the user clicks this button, a modal will pop up that will show important information regarding the draft artifact such as its name, status, id, url, description, and effective period. If a draft artifact doesn’t have one of these properties defined, they will not appear in the modal. At the bottom of the modal, the user will be required to enter in a version of the draft before they can release it. Once they add the version and click the “Release” button, the draft version of the artifact will be deleted from the authoring portion of the repository and a new version of the artifact will be “released” to the repository. This new artifact will have an active status and whatever version the user added. 

## Code changes
`app/src/components/DraftListItem.tsx` —> This creates a customized listItem component for a list. 
`app/src/components/ReleaseModal.tsx` —> This creates a customized modal that contains all important information regarding the artifact
`confirm()` function in `app/src/components/ReleaseModal.tsx` —> Function that “releases” the draft artifact

# Testing guidance
```
npm run check:all
npm run start:all
```
Navigate to the edit page for a draft artifact in the authoring portion of the MRS. Test with draft artifacts that were created from existing artifacts and ones that were just created. Make sure the “Release button” is rendered for each artifact in the edit page.
- Click on the release button and make sure the modal appears. Check the modal for the following:
    - If the artifact has a name, status, id, url, type, description, or effective period, then these properties should appear on the modal. Otherwise they should not appear at all. 
    - The different properties should appear with alternating background colors and when the user hovers over a specific property, its corresponding div element should highlight.
    - The release button on the modal should be disabled until the user enters a version for the artifact
- After entering a version for the artifact, click the release button. A success notification should pop up and  then you will be routed to the details page of the new released artifact in the repository section. Check the following:
    - On the details page of the new released artiface, make sure the version you entered was added properly.
    - In the authoring section, make sure that the draft artifact button for the given resource type is decremented by 1
    - In the authoring section, make sure that the resource’s card no longer appears on the page
    - Also make sure the artifact no longer exists in the draft portion of the database and now exists in the actual repository in studio3T

Lauren left some great comments/questions scattered throughout the code that I left because I was uncertain about their answers and thought it would require some more consideration in the future.

### DIFF
--- a/app/src/components/DraftListItem.tsx
+++ b/app/src/components/DraftListItem.tsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { Center, Grid, List, Space, Table, Text } from '@mantine/core';
+import { useEffect, useState } from 'react';
+import { useHover } from '@mantine/hooks';
+
+interface modalProps {
+  header: string;
+  context?: string;
+  originalColor: string;
+  date?: any;
+}
+
+export default function DraftListItem({ header, context, originalColor, date }: modalProps) {
+  const [color, setColor] = useState(originalColor);
+  const { hovered, ref } = useHover();
+
+  // Changes the color of the div element text depending on if the user is hovering over it
+  useEffect(() => {
+    if (hovered === true) {
+      setColor('#E9ECEF');
+    } else {
+      setColor(originalColor);
+    }
+  }, [hovered, originalColor]);
+
+  if (date) {
+    const rows = (
+      <tr key={date}>
+        <td>{date.start}</td>
+        <td>{date.end}</td>
+      </tr>
+    );
+    const ths = (
+      <tr>
+        <th>{`Start Date`}</th>
+        <th>{`End Date`}</th>
+      </tr>
+    );
+
+    return (
+      <>
+        <Center>
+          <List.Item>
+            <div
+              ref={ref}
+              style={{
+                width: '900px',
+                backgroundColor: color,
+                height: '95px',
+                wordBreak: 'break-word',
+                overflow: 'scroll'
+              }}
+            >
+              <Grid justify="apart">
+                <Grid.Col span={3}>
+                  <Center h={'75px'}>
+                    <Text size="md" color="#495057" fw={400}>
+                      <b>{header}</b>:
+                    </Text>
+                  </Center>
+                </Grid.Col>
+                <Center h={'115px'}>
+                  <List.Item>
+                    <Table horizontalSpacing="xl" highlightOnHover withBorder withColumnBorders>
+                      <thead>{ths}</thead>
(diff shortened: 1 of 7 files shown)


app/src/components/DraftListItem.tsx
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
+import { Center, Grid, List, Space, Table, Text } from '@mantine/core';
+import { useEffect, useState } from 'react';
+import { useHover } from '@mantine/hooks';
+
+interface modalProps {
```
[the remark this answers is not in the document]
elsaperelli: Actually, since this isn't a modal but just part of the modal, I would call it `DraftListItemProps`

app/src/components/ReleaseModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  //This function is needed because, if I were to include this functionality where I call it,
+  // then it would appear on the modal which I don't want to happen
+  const setColor = (newColor: string) => {
+    currentColor = newColor;
+  };
+
+  return (
+    <Modal opened={open} onClose={onClose} withCloseButton={false} size="65%" centered>
```
elsaperelli, line 122: One thing I noticed which I should have brought up sooner is that sometimes UI/UX can be a pain when it comes to resizing the window. We don't need to worry about every window size, but it's good practice to enlargen or shrink your window to make sure the UI looks somewhat doable at different sizes. For example, this is what the modal current looks like when my window is a bit smaller than my laptop screen (which I think is a reasonable size for users:

image

I believe that the `size="65%"` here is partially the culprit, but in my experience, this requires some trial and error.

elsaperelli (changes requested):
Woo last PR of your internship! I know this was Lauren's task too but I love what you did with the UI here. 

I have some small comments but my main one which is also my suggestion for something do play around with in the time you have left is working on the sizing of things. It looks like there are a lot of sizes, heights, etc. that are hardcoded in the React components. We try to avoid that as much as possible since we want to account for varying window sizes. See you if you change some of these and test different window sizes!

In terms of the questions that Lauren left, I am holding off on responding to them for now. I think a lot of these are questions that we will have to investigate more deeply in the future.

app/src/components/ReleaseModal.tsx
BriannaVH: Great catch! I didn't realize just how much I hard-coded sizing into the modal. I added some checks on the modal itself but not the list items and I was wondering if this helped your sizing in any way?

PR #59, opened 2023-10-04
Cache data-requirements

elsaperelli:
# Summary
When the `$data-requirements` for a specific artifact and set of parameters are calculated, they are then cached in the mongo database that stores active artifacts so that these Library data requirements resources can be accessed again without having to rerun calculation. 

## New behavior
Before, whenever the user requested the data-requirements for an artifact (either with the `$data-requirements` endpoint or in the UI, the data-requirements calculation was done each time. In order to improve performance in our application, we now cache these Library resources in the Library collection in the published repository service.

## Method:
We talked about different strategies for how to cache data-requirements and where but ultimately settled on the following (for now):
1. When the user navigates to an artifact's page in the UI, `$data-requirements` is called (same behavior as before). 
2. Before, this endpoint would _only_ calculate the data-requirements of a Library or Measure and simply return them. Now, we look in the Library collection to see if the data-requirements of the requested artifact and associated parameters have been calculated before and if so return them. If not, they are calculated and cached in the Library collection and the data-requirements are still returned.
3. We also decided to not calculate data-requirements for an artifact when it is released since the application will route to the released artifacts page which already calls `$data-requirements`.

## Code changes
- `app/src/pages/[resourceType]/[id].tsx` - On the main branch, you will notice some odd behavior in the relatedArtifacts in the Data Requirements tab of an artifact. The previous code was sorting the relatedArtifacts in place. I pulled the sortDependencies out to a function and this gets called on a copy of the relatedArtifacts of data requirements so that the relatedArtifacts are not sorted in place. The sort is only called when the dataRequirements value changes.
- `app/src/pages/authoring/[resourceType]/[id].tsx` - Removed comment
- `package-lock.json` - this changed when I did `npm install --workspace=app --workspace=service`...I think this is okay?
- `dbOperations.ts` - `findResourceById` and `findResourcesWithQuery` make sure to exclude Libraries that have `_dataRequirements` property (they are Library data requirements resources). `findDataRequirementswithQuery` returns the Library data requirements resource that is associated with an artifact and set of query parameters if it exists.
- `LibraryService.ts` / `MeasureService.ts` - `$data-requirements` endpoint now checks mongo collection for cached data requirements and returns it if it exists. If not, it calculates the data-requirements, caches them, and returns them.
- `service/src/types/service-types.ts` - new type for Library data requirements resources to additionally contain an optional `_dataRequirements` property that contains the query associated with the data requirements call.
- `LibraryService.test.ts` / `MeasureService.test.ts` - When `calc` was not reset each time, I was getting a mongo duplicate key error. 

# Testing guidance
- `npm run check:all`
- `npm run start:all` 

Things to look for:
- Functionality for Dependencies tab / search are the same.
- Data requirements is only calculated once for an artifact (you should see a bunch of logs for assembling the data requirements once and then after that only the message that they were successfully retrieved).
- Make sure that the data requirements Library resources are not displayed in any of the Library artifact lists.

One more thing:
- The data requirements output when it is first calculated is slightly different from the data requirements output when it is grabbed from the cache. This is due to `version` being undefined in the output from `fqm-execution`, so when it gets put in mongo it becomes "version"" null. I made that quick code change in mongo and did `npm link` to this repository to confirm that that was the issue and it is. So whenever that is fixed in `fqm-execution`, the outputs will be the same. For now, I am going to make a task to further investigate this in `fqm-execution` and hopefully address it.

### DIFF
--- a/service/src/services/LibraryService.ts
+++ b/service/src/services/LibraryService.ts
@@ -1,12 +1,12 @@
 import { loggers, RequestArgs, RequestCtx } from '@projecttacoma/node-fhir-server-core';
 import {
+  createResource,
   findDataRequirementsWithQuery,
   findResourceById,
   findResourcesWithQuery,
   updateResource
 } from '../db/dbOperations';
 import { LibrarySearchArgs, LibraryDataRequirementsArgs, PackageArgs, parseRequestSchema } from '../requestSchemas';
-import { createResource } from '../db/dbOperations';
 import { Service } from '../types/service';
 import { createLibraryPackageBundle, createSearchsetBundle } from '../util/bundleUtils';
 import { BadRequestError, ResourceNotFoundError } from '../util/errorUtils';
@@ -22,7 +22,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { Calculator } from 'fqm-execution';
 const logger = loggers.get('default');
 import { Filter } from 'mongodb';
-import { FhirResourceWithDR } from '../types/service-types';
+import { FhirLibraryWithDR } from '../types/service-types';
 
 /*
  * Implementation of a service for the `Library` resource
@@ -158,7 +158,7 @@ export class LibraryService implements Service<fhir4.Library> {
     });
 
     dataRequirements.results['id'] = uuidv4();
-    const results = { ...dataRequirements.results } as FhirResourceWithDR;
+    const results = { ...dataRequirements.results } as FhirLibraryWithDR;
 
     // add the data requirements query params to the data requirements Library resource and add to the Library collection
     results['_dataRequirements'] = dataReqsQuery;
--- a/service/src/services/LibraryService.ts
+++ b/service/src/services/LibraryService.ts
@@ -158,9 +158,9 @@ export class LibraryService implements Service<fhir4.Library> {
     });
 
     dataRequirements.results['id'] = uuidv4();
-    const results = { ...dataRequirements.results } as FhirLibraryWithDR;
 
     // add the data requirements query params to the data requirements Library resource and add to the Library collection
+    const results = { ...dataRequirements.results } as FhirLibraryWithDR;
     results['_dataRequirements'] = dataReqsQuery;
     createResource(results, 'Library');
 
--- a/service/src/services/LibraryService.ts
+++ b/service/src/services/LibraryService.ts
@@ -1,5 +1,10 @@
 import { loggers, RequestArgs, RequestCtx } from '@projecttacoma/node-fhir-server-core';
-import { findResourceById, findResourcesWithQuery, updateResource } from '../db/dbOperations';
+import {
+  findDataRequirementsWithQuery,
+  findResourceById,
+  findResourcesWithQuery,
+  updateResource
+} from '../db/dbOperations';
 import { LibrarySearchArgs, LibraryDataRequirementsArgs, PackageArgs, parseRequestSchema } from '../requestSchemas';
 import { createResource } from '../db/dbOperations';
 import { Service } from '../types/service';
@@ -16,6 +21,8 @@ import {
 import { v4 as uuidv4 } from 'uuid';
 import { Calculator } from 'fqm-execution';
 const logger = loggers.get('default');
+import { Filter } from 'mongodb';
+import { FhirResourceWithDR } from '../types/service-types';
 
 /*
  * Implementation of a service for the `Library` resource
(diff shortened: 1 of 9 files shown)


service/src/services/LibraryService.ts
```diff
@@ -124,15 +131,40 @@ export class LibraryService implements Service<fhir4.Library> {
...
+    });
+
+    const dataReqs = await findDataRequirementsWithQuery(dataReqsQuery);
+
+    // if data requirements were already calculated for this Library and params, return them
+    if (dataReqs) {
+      logger.info('Successfully retrieved $data-requirements report');
+      return dataReqs;
+    }
+
```
sarahmcdougall, line 151: As we build out the data requirements functionality between the Library and Measure service, it looks like we are reusing a lot of code. Do you think it's possible to pull some of this out into a function that can be used by both services?

service/src/services/LibraryService.ts
```diff
@@ -124,15 +131,40 @@ export class LibraryService implements Service<fhir4.Library> {
...
+      } else {
+        dataReqsQuery[p[0]] = p[1] as string;
+      }
+    });
+
+    const dataReqs = await findDataRequirementsWithQuery(dataReqsQuery);
+
+    // if data requirements were already calculated for this Library and params, return them
+    if (dataReqs) {
+      logger.info('Successfully retrieved $data-requirements report');
```
sarahmcdougall, line 148:
```suggestion
      logger.info('Successfully retrieved $data-requirements report from cache.');
```
Nitpick but adding some more context makes it easier to understand whether the data requirements were cached or if they are being calculated for the first time

service/src/services/MeasureService.ts
```diff
@@ -1,5 +1,5 @@
 import { loggers, RequestArgs, RequestCtx } from '@projecttacoma/node-fhir-server-core';
-import { findResourceById, findResourcesWithQuery } from '../db/dbOperations';
+import { findDataRequirementsWithQuery, findResourceById, findResourcesWithQuery } from '../db/dbOperations';
```
sarahmcdougall, line 2:
```suggestion
import { findDataRequirementsWithQuery, findResourceById, findResourcesWithQuery, createResource, updateResource } from '../db/dbOperations';
```
Same suggestion that was mentioned in LibraryService - can add the rest of the imports here

service/src/services/MeasureService.ts
```diff
@@ -15,6 +15,8 @@ import { Calculator } from 'fqm-execution';
 import { MeasureSearchArgs, MeasureDataRequirementsArgs, PackageArgs, parseRequestSchema } from '../requestSchemas';
 import { v4 as uuidv4 } from 'uuid';
 import { createResource, updateResource } from '../db/dbOperations';
```
sarahmcdougall, line 17:
```suggestion
```

service/src/services/LibraryService.ts
```diff
@@ -124,15 +131,40 @@ export class LibraryService implements Service<fhir4.Library> {
     const query = extractIdentificationForQuery(args, params);
     const parsedParams = parseRequestSchema({ ...params, ...query }, LibraryDataRequirementsArgs);
 
+    // check to see if data requirements were already calculated for this Library and params
+    const dataReqsQuery: Filter<any> = {};
+    Object.entries(parsedParams).forEach(p => {
```
sarahmcdougall, line 136: Just checking my understanding of what we want to query for here... we don't necessarily need to loop over all the `parsedParams`, right? We are really just interested in the identifying parameters (`id`, `url`, `version`, `identifier`) in terms of the query itself? Looks like `parsedParams` can include fields like `check-system-vesrion` and `include-dependencies`, but doesn't look like their inclusion in the data reqs query has any effect on the result

sarahmcdougall (changes requested):
Good work!
I made a placeholder task in the backlog to further investigate the undefined version issue that you mentioned in the PR description: https://jira.mitre.org/browse/TACOSTRAT-1398

Aside from in-line comments, I noticed that the app is especially slow when testing on a library artifact. I think this is because all the dependent libraries get retrieved first and then the data requirements are calculated, so it seems out of scope for this task. 

I got this error in the terminal, which I don't believe I have gotten before:
Screenshot 2023-10-09 at 2 55 28 PM

I documented this error as a comment on the Jira ticket for investigating performance issues

hossenlopp (changes requested):

service/src/services/LibraryService.ts
elsaperelli: Hmm. I see your point and for the most part agree, but I think I am going to defer to @hossenlopp on this one since I know we want to be specific when it comes to the parameters used.

sarahmcdougall (approved):
The changes look good! Good work on this!

service/src/services/LibraryService.ts
hossenlopp: This should work for now and set us up to work with additional params.

hossenlopp (approved):

PR #65, opened 2023-11-06
Update package endpoint to cqfm.package

sarahmcdougall:
# Summary
The [published version of the measure repository service](http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package) now specifies `$cqfm.package` as the endpoint for the package operation. This PR changes instances of `$package` with `$cqfm.package`.

## New behavior
No new features. Changes the package endpoint from `$package` to `$cqfm.package`. Also updates our schema instances, which impacts which parameters throw errors when caught being used without support from the server.

## Code changes
* README updates
* `capabilityStatementResources.json` - update OperationDefinition link for the package operation
* `serverConfig.ts` - update operation endpoint and reference
* `requestSchemas.ts` - Made updates to unsupported package args and general package args schema based on changes to the spec.
* Updates to `LibraryService` and `MeasureService` tests (update the mock requests)
* Other miscellaneous changes to update documentation related to the change in endpoint

# Testing guidance
* Run unit tests for the measure repository service and ensure that they all pass
* Run various`$cqfm.package` requests - The Insomnia imports from previous pull requests can be used as a starting point for making these requests (the only change that needs to be made within these imports is the change from `$package` to `$cqfm.package`). All successful requests should return `200` status and a FHIR Bundle.
    * Example endpoints include `[base]/Library/$cqfm.package`, `[base]/Library/[id]/$cqfm.package`, etc.
    * Test both POST and GET requests
* Check that the request schema parameters align with the parameters listed in the measure repository service definition. There are a few inconsistencies between the [CQFM Package OperationDefinition](http://hl7.org/fhir/us/cqfmeasures/STU4/OperationDefinition-cqfm-package.html), the [CQFM Package section](http://hl7.org/fhir/us/cqfmeasures/measure-repository-service.html#package) of the measure repository page, and the [Publishable Measure Repository Section](http://hl7.org/fhir/us/cqfmeasures/measure-repository-service.html#libraries-1). Check that the TODO comments align with the inconsistencies that need to be addressed.

Insomnia tests are included below (endpoint will need to be changed to `$cqfm.package`):
[Library-$package-testing.json](https://github.com/projecttacoma/measure-repository/files/13270438/Library-.package-testing.json)
[Measure-$package-testing.json](https://github.com/projecttacoma/measure-repository/files/13270441/Measure-.package-testing.json)

### DIFF
--- a/service/src/config/capabilityStatementResources.json
+++ b/service/src/config/capabilityStatementResources.json
@@ -140,7 +140,7 @@
               "valueCode": "SHALL"
             }
           ],
-          "name": "package",
+          "name": "cqfm-package",
           "definition": "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package"
         },
         {
--- a/service/src/config/capabilityStatementResources.json
+++ b/service/src/config/capabilityStatementResources.json
@@ -141,7 +141,7 @@
             }
           ],
           "name": "package",
-          "definition": "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Library-package"
+          "definition": "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package"
         },
         {
           "extension": [
@@ -297,7 +297,7 @@
             }
           ],
           "name": "package",
-          "definition": "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/Measure-package"
+          "definition": "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package"
         },
         {
           "extension": [
--- a/service/src/config/capabilityStatementResources.json
+++ b/service/src/config/capabilityStatementResources.json
@@ -296,7 +296,7 @@
               "valueCode": "SHALL"
             }
           ],
-          "name": "package",
+          "name": "cqfm-package",
           "definition": "http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package"
         },
         {
--- a/service/src/config/serverConfig.ts
+++ b/service/src/config/serverConfig.ts
@@ -39,25 +39,25 @@ export const serverConfig: ServerConfig = {
       versions: [constants.VERSIONS['4_0_1']],
       operation: [
         {
-          name: 'package',
+          name: 'cqfm.package',
           route: '/$cqfm.package',
           method: 'GET',
           reference: 'http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package'
         },
         {
-          name: 'package',
+          name: 'cqfm.package',
           route: '/$cqfm.package',
           method: 'POST',
           reference: 'http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package'
         },
         {
-          name: 'package',
+          name: 'cqfm.package',
           route: '/:id/$cqfm.package',
           method: 'GET',
           reference: 'http://hl7.org/fhir/us/cqfmeasures/OperationDefinition/cqfm-package'
         },
         {
-          name: 'package',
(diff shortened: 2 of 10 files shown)


service/src/config/capabilityStatementResources.json
```diff
@@ -141,7 +141,7 @@
             }
           ],
           "name": "package",
```
elsaperelli, line 143: Do we want to change the name here? I am not sure so I thought I'd ask

service/src/config/serverConfig.ts
```diff
@@ -40,27 +40,27 @@ export const serverConfig: ServerConfig = {
       operation: [
         {
           name: 'package',
```
elsaperelli, line 42: Same here?

service/src/requestSchemas.ts
```diff
@@ -117,15 +121,20 @@ export const IdentifyingParameters = z
...
+  canonicalVersion: z.string(),
   capability: z.string(),
-  'check-system-version': z.string(),
+  contentEndpoint: z.string(),
+  checkCanonicalVersion: z.string(),
   count: stringToNumber,
-  'force-system-version': z.string(),
-  'include-components': stringToBool,
-  'include-dependencies': stringToBool,
+  forceCanonicalVersion: z.string(),
```
elsaperelli, line 129: Do we want a TODO comment for the `canonicalVersion` vs. `system-version` params too?

elsaperelli (changes requested):
Small comments, otherwise looks good!

service/src/config/capabilityStatementResources.json
sarahmcdougall: Yes, for consistency I think it's a good idea to update this!

service/src/config/serverConfig.ts
sarahmcdougall: Tried making the change here and it made all the unit tests fail. I forgot that the [asymmetrik docs] require that the name matches up with the corresponding function name, where the function name is camel case. Right now the name is `"package"` and the corresponding function is `package()`. If we were to change the name, we would need to change the function to `cqfmPackage()`. It's not a big deal if we were to add that change, but I think it's fine to keep this as just `package` for now since there is only one package function defined within the measure repository spec anyway.

elsaperelli: Ohhh oops! Sorry about that, I totally forgot about the asymmetrik stuff. No need to change!

service/src/config/capabilityStatementResources.json
elsaperelli: Honestly maybe because we are keeping the `package` name in `serverConfig.ts` should we change this back? I am sorry about all this hahah

elsaperelli (approved):
Lgtm! Sorry about all the rename craziness haha, up to you whether to change back `cqfm.package` to just `package` in `capabilityStatementResources.json`. Other than that, it's good to go!

sarahmcdougall:
> Lgtm! Sorry about all the rename craziness haha, up to you whether to change back `cqfm.package` to just `package` in `capabilityStatementResources.json`. Other than that, it's good to go!

I think it's fine to keep it as `cqfm.package` - the `name` field in the capability statement doesn't really impact anything, unlike the name for the server config.

service/src/config/capabilityStatementResources.json
```diff
@@ -297,7 +297,7 @@
...
```
hossenlopp, line 299:
```suggestion
          "name": "cqfm-package",
```
I think this one should be `cqfm-package` too.

hossenlopp (changes requested):
Looks like just one little thing in CapabilityStatement needs updating.

hossenlopp (approved):

PR #69, opened 2023-11-28
Child artifact support

elsaperelli:
# Summary
This PR adds the majority of the child artifacts support in the authoring operations in the `app` side of this repository as well as adds the `isOwned` extension to child artifacts.

## New behavior
The first part of this task involved adding the [`isOwned` extension](https://build.fhir.org/ig/HL7/crmi-ig/branches/master/StructureDefinition-artifact-isOwned.html) to the child artifact in the `service`/"backend" of the measure repository. When the user uses our `dbSetup.ts` script (`db:loadBundle`) for uploading bundles, the following happens to the measure artifact in the bundle to ensure child artifacts will be supported in the authoring operations in the application:
- The main library and version of the measure artifact are concatenated for the url to the following format: mainLibrary|version. If there is no version, then just the mainLibrary is used.
- If the measure artifact does not contain a `relatedArtifacts` array, then one is created.
- The mainLibrary and the `isOwned` extension are added as an entry to the `relatedArtifacts` array in the following structure:

```
"relatedArtifact": [
    {
        "type": "composed-of",
        "resource": <mainLibrary|version>,
        "extension": [
            {
                "url": "http://hl7.org/fhir/StructureDefinition/artifact-isOwned",
                "valueBoolean": true
            }
...
```

The second part of this task involves the actual child artifact handling in the `draft` and the `release` authoring operations ONLY. 

For the [draft](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#draft) operation:
- Call the `draftParent` mutation on the artifact to be drafted
- `getChildren` is called on the artifact's relatedArtifacts in order to recursively get all of the child artifacts on the artifact, if they exist
- The artifact is converted to a draft through `modifyResourceToDraft`
- The newly modified artifact is created as a draft on the authoring database
- The new draft artifact's id and the array child artifacts is returned
- If the `draftParent` mutation was successful, then the `draftChild` mutation is called on each of the child artifacts

For the [release](https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#release) operation:
- Call the `releaseParent` mutation on the draft artifact to be released
- The draft artifact is modified to have an active status and is then released to the published measure repository
- Get the location of the newly released artifact 
- `getChildren` is called on the artifact's relatedArtifacts in order to recursively get all of the child artifacts on the artifact, if they exist
- The newly released artifact's location, status, and the array of child artifacts is returned
- If the `releaseParent` mutation was successful, then...
- The parent draft artifact gets deleted from the draft database
- The `releaseChild` mutation is called on each of the child artifacts
- If the `releaseChild` mutation is successful, then the child draft artifact gets deleted from the draft database

## Code changes
- `app/src/components/ReleaseModal.tsx` - includes two mutations: `releaseMutation` and `releaseChildMutation`, move success notification from delete mutation to release mutation 
- `app/src/pages/[resourceType]/[id].tsx` - make success and error notifications reusable, includes two mutations: `draftFromArtifactMutation` and `draftChildMutation`
- `app/src/pages/authoring/index.tsx` - make success and error notifications reusable, includes two mutations: `draftFromArtifactMutation` and `draftChildMutation`
- `app/src/server/db/dbOperations.ts` - adds `getDraftByUrl` function that retrieves the draft resource of the given type with the given url and version.
- `app/src/server/trpc/routers/draft.ts` - added `getDraftByUrl` trpc procedure
- `app/src/server/trpc/routers/service.ts` - added a helper function `getChildren` to be used in two procedures, added four new procedures: `draftParent`, `draftChild`, `releaseParent`, and `releaseChild`
- `dbSetup.ts` - add `isOwned` extension to `relatedArtifact` on artifacts when they are uploaded to the server using the script
- `app/src/util/modifyResourceFields.ts` - increments the version on a child artifact's `resource`
- `app/src/util/types/fhir.ts` - add `id` (server-defined) to `CQFMMeasure` and `CQFMLibrary` interfaces

# Testing guidance
- `npm run check:all`

Reset the measure repository and re upload bundles
- `cd service` 
- `npm run db:reset`
- `npm run db:loadBundle <path to bundle>` (I tested with the ColorectalCancerScreeningsFHIR bundle

Spin up the server and frontend app:
- In the root directory: `npm run start:all`
- Look at the measure that you just uploaded to the server when uploading the bundle. The measure should have a "relatedArtifact" property that is an array that is populated with the main library of the measure. There should also be the "isOwned" extension on this entry.
- Test drafting existing artifacts and releasing draft artifacts. Keep in mind that these were the only two authoring operations that I added child artifact support for.

Things to note:
- Since the measure repository service spec is often not clear, there were some assumptions/decisions that I had to make regarding the process. I think for the most part I follow the spec, but include that in your review of this PR: make sure the functionality aligns with the spec as much as possible.

### DIFF
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -1,11 +1,9 @@
 import { trpc } from '@/util/trpc';
 import { ArtifactResourceType } from '@/util/types/fhir';
 import { Button, Center, Group, Modal, Stack, Text, Tooltip } from '@mantine/core';
-import { DateTime } from 'luxon';
 import { useRouter } from 'next/router';
 import { AlertCircle, CircleCheck, InfoCircle } from 'tabler-icons-react';
 import { notifications } from '@mantine/notifications';
-import { release } from 'os';
 
 export interface ReleaseModalProps {
   open: boolean;
@@ -44,9 +42,9 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
     }
   });
 
-  const releaseMutation = trpc.service.releaseChildren.useMutation({
+  const releaseMutation = trpc.service.releaseArtifactByUrl.useMutation({
     onSuccess: data => {
-      if (!data[0].location) {
+      if (!data.location) {
         console.error('No resource location for released artifact');
         notifications.show({
           title: `Release Failed!`,
@@ -54,8 +52,8 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
           icon: <AlertCircle />,
           color: 'red'
         });
-      } else if (data[0].res.status !== 201) {
-        console.error(data[0].res.status);
+      } else if (data.status !== 201) {
+        console.error(data.status);
         notifications.show({
           title: `Release Failed!`,
           message: `Server unable to process request`,
@@ -63,7 +61,7 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
           color: 'red'
         });
       } else {
-        router.push(data[0].location);
+        router.push(data.location);
         deleteMutation.mutate({
           resourceType: resourceType,
           id: id
@@ -78,15 +76,12 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
     // https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#release
     // TODO: release recursively all children (ignore for now).
     if (resource) {
-      resource.status = 'active';
-      resource.date = DateTime.now().toISO() || '';
+      releaseMutation.mutate({
+        resourceType: resourceType,
+        id: id,
+        version: resource.version
+      });
     }
-
-    releaseMutation.mutate({
-      resourceType: resourceType,
-      id: id,
-      version: version
-    });
   }
 
   return (
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -61,7 +61,7 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
       console.error(e);
       notifications.show({
         title: `Release Failed!`,
-        message: `Attempt to release ${variables.resourceType} failed with message: ${e.message}`,
+        message: `Attempt to release child ${variables.resourceType} with url ${variables.url} failed with message: ${e.message}`,
         icon: <AlertCircle />,
         color: 'red'
       });
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -20,11 +20,12 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
     resourceType: resourceType
   });
   const ctx = trpc.useContext();
+
   const deleteMutation = trpc.draft.deleteDraft.useMutation({
-    onSuccess: () => {
+    onSuccess: data => {
       notifications.show({
-        title: `Draft ${resource?.resourceType} released!`,
-        message: `Draft ${resource?.resourceType}/${resource?.id} successfully released to the Publishable Measure Repository!`,
+        title: `Draft ${data.resourceType} released!`,
+        message: `Draft ${data.resourceType}/${data.draftId} successfully released to the Publishable Measure Repository!`,
         icon: <CircleCheck />,
         color: 'green'
       });
@@ -35,7 +36,7 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
       console.error(e);
       notifications.show({
         title: `Release Failed!`,
-        message: `Attempt to release ${resourceType} failed with message: ${e.message}`,
+        message: `Attempt to release artifact failed with message: ${e.message}`,
         icon: <AlertCircle />,
         color: 'red'
       });
@@ -65,6 +66,12 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
         deleteMutation.mutate({
           resourceType: resourceType,
           id: id
+        });
+        data.children.forEach(childArtifact => {
+          deleteMutation.mutate({
+            resourceType: childArtifact.resourceType,
+            id: childArtifact.id
+          });
         });
       }
       onClose();
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -43,6 +43,35 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
     }
   });
 
+  const releaseMutation = trpc.service.releaseArtifactById.useMutation({
+    onSuccess: data => {
+      if (!data.location) {
+        console.error('No resource location for released artifact');
+        notifications.show({
+          title: `Release Failed!`,
+          message: `No resource location exists for draft artifact`,
+          icon: <AlertCircle />,
+          color: 'red'
+        });
+      } else if (data.status !== 201) {
+        console.error(data.status);
+        notifications.show({
+          title: `Release Failed!`,
+          message: `Server unable to process request`,
+          icon: <AlertCircle />,
+          color: 'red'
+        });
+      } else {
+        router.push(data.location);
+        deleteMutation.mutate({
+          resourceType: resourceType,
+          id: id
+        });
+      }
+      onClose();
+    }
+  });
+
   async function confirm() {
     // requirements:
     // https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#release
@@ -52,47 +81,11 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
       resource.date = DateTime.now().toISO() || '';
     }
 
-    const res = await fetch(`${process.env.NEXT_PUBLIC_MRS_SERVER}/${resourceType}`, {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json+fhir'
-      },
-      body: JSON.stringify(resource)
+    releaseMutation.mutate({
+      resourceType: resourceType,
+      id: id,
+      version: version
     });
-
-    // Conditionally remove the base version if one is present
-    let location = res.headers.get('Location');
-    if (location?.substring(0, 5) === '4_0_1') {
-      location = location?.substring(5); // remove 4_0_1 (version)
-    }
-
-    if (res.status !== 201) {
-      console.error(res.statusText);
-      notifications.show({
-        title: `Release Failed!`,
-        message: `Server unable to process request`,
-        icon: <AlertCircle />,
-        color: 'red'
-      });
-    } else if (!location) {
-      console.error('No resource location for released artifact');
-      notifications.show({
-        title: `Release Failed!`,
-        message: `No resource location exists for draft artifact`,
-        icon: <AlertCircle />,
-        color: 'red'
-      });
-    } else {
-      // delete draft
-      deleteMutation.mutate({
-        resourceType: resourceType,
-        id: id
-      });
-
-      // direct user to published artifact detail page
-      router.push(location);
-    }
-    onClose();
   }
 
   return (
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -5,6 +5,7 @@ import { DateTime } from 'luxon';
 import { useRouter } from 'next/router';
 import { AlertCircle, CircleCheck, InfoCircle } from 'tabler-icons-react';
 import { notifications } from '@mantine/notifications';
+import { release } from 'os';
 
 export interface ReleaseModalProps {
   open: boolean;
@@ -43,9 +44,9 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
     }
   });
 
-  const releaseMutation = trpc.service.releaseArtifactById.useMutation({
+  const releaseMutation = trpc.service.releaseChildren.useMutation({
     onSuccess: data => {
-      if (!data.location) {
+      if (!data[0].location) {
         console.error('No resource location for released artifact');
         notifications.show({
           title: `Release Failed!`,
@@ -53,8 +54,8 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
           icon: <AlertCircle />,
           color: 'red'
         });
-      } else if (data.status !== 201) {
-        console.error(data.status);
+      } else if (data[0].res.status !== 201) {
+        console.error(data[0].res.status);
         notifications.show({
           title: `Release Failed!`,
           message: `Server unable to process request`,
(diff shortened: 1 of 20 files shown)


sarahmcdougall (changes requested):
Aside from in-line comments, I am having some trouble with testing the following scenarios:
1. When I am viewing an active artifact and click the "Create Draft of Measure" button, a draft is created for the measure,  but no draft is created for the child artifact. 
Screenshot 2023-11-30 at 1 58 10 PM

However, I am able to successfully draft a measure and its child when I do so from "authoring" mode:
Screenshot 2023-11-30 at 1 58 55 PM

I would suggest incorporating functionality for child artifacts to be drafted from option 1.

2. When I release an artifact, the artifact and its child are successfully released. However, only the parent artifact gets deleted. The (draft) child artifact does not get deleted. I would suggest calling `deleteMutation.mutate` on the children - it looks like we should have access to the child from the release modal, so I don't think this should cause too much overhead.

app/src/server/trpc/routers/service.ts
```diff
@@ -1,9 +1,122 @@
...
+      )
+    ) {
+      let raType: ArtifactResourceType;
+      if (ra.resource?.includes('Measure')) {
+        raType = 'Measure';
+      } else {
+        raType = 'Library';
+      }
+
+      const [url, version] = ra.resource.split('|');
```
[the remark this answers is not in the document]
elsaperelli: Good call!

app/src/server/db/dbOperations.ts
```diff
@@ -19,6 +19,19 @@ export async function getDraftById<T extends FhirArtifact>(id: string, resourceT
   return collection.findOne<T>({ id }, { projection: { _id: 0 } });
 }
 
+/**
+ * Retrieves the draft resource of the given type with the given url and version
+ */
+export async function getDraftByUrl<T extends FhirArtifact>(
+  url: string,
+  version: string,
```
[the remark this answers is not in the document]
elsaperelli: Happy to discuss this in more detail, but short answer is yes. This function is getting a draft artifact (from the draft artifact database) by its url. We determined that drafts must have a version; therefore, whenever we create a draft artifact, we ensure that it has a version (either by incrementing the existing artifact's version or giving it a version of 0.0.1). I am sure the version is optional in the zod validation because we used to define FhirArtifact as a Measure or Library but they should actually be a CQFMMeasure or CQFMLibrary whose version has cardinality of 1..1. Should I look into this further in this task? I am trying my best to not let things get too out of hand, so maybe we should create a task for further refining typing based on new conclusions?

app/src/server/trpc/routers/service.ts
```diff
@@ -1,9 +1,122 @@
...
+        body: JSON.stringify(draftRes)
+      });
+
+      let location = res.headers.get('Location');
+      if (location?.substring(0, 5) === '4_0_1') {
+        location = location?.substring(5); // remove 4_0_1 (version)
+      }
+
+      // add response to array
+      result.push({ resourceType: raType, id: url, res: res, location: location });
```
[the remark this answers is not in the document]
elsaperelli: I believe I resolved this with the re-working of changing the fetch to use PUT instead of POST, but let me know!

app/src/server/trpc/routers/service.ts
```diff
@@ -1,9 +1,122 @@
...
+  res: Response;
+  location: string | null;
+}
+
+export type ArtifactInfo = {
+  resourceType: 'Measure' | 'Library';
+  id: string;
+};
+
+async function draftChildren(relatedArtifacts: fhir4.RelatedArtifact[]) {
```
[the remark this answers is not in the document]
elsaperelli: I like this idea! What would the most helpful logs look like? We log whenever a draft is created or released (including child artifacts), what else would you like to see included?

elsaperelli:
@sarahmcdougall Good catch on your first point! I think that should be all good now. Regarding your second point, good call that we can easily call `deleteMutation.mutate` on the children. I initially didn't do this because I was going to do the recursive delete in another task and just focus on recursion for `draft` and `release` for this task.  I agree that this was an easy enough add; however, in the future I may be changing the entire `deleteMutation` procedure to use recursion, so this may change but I think that's okay. I also realized that the server defined id for draft artifacts was changing too much so I changed the POST request for releasing a draft artifact to be a PUT request with the draft artifact id. Before, when we drafted an artifact, a new id was created AND when that draft artifact was then released. I believe we want the id to stay the same when it goes from draft to active artifact since we already created a new id when the artifact was drafted. Let me know if that makes sense!

service/scripts/dbSetup.ts
```diff
@@ -49,6 +49,53 @@ async function loadBundle(filePath: string) {
...
+
+          if (mainLibraryRA) {
+            // check if the main library's extension array exists and create it if it doesn't
+            if (mainLibraryRA.extension) {
+              // if it does exist, check that the isOwned extension is not already on it, add it if not
+              if (
+                mainLibraryRA.extension.find(
+                  e => e.url === 'http://hl7.org/fhir/StructureDefinition/artifact-isOwned' && e.valueBoolean === true
+                ) === undefined
+              )
```
sarahmcdougall, line 80: It might make more sense here to not include the `e.valueBoolean === true` case in the `.find`. I'm thinking of an edge case where an artifact has specifically been marked to *not* be an owned artifact, meaning that (1) the extension is present, and (2) `valueBoolean` is explicitly set to `false`. If this were the case, would we then have two extension entries, one with `valueBoolean` set to `true` and `valueBoolean` set to `false`?

app/src/server/db/dbOperations.ts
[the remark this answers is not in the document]
sarahmcdougall: Thanks for the explanation! I forgot about the intricacies of the zod validation in regard to optional fields. I think it's fine to leave the current functionality for now.

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -84,30 +83,45 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
+      : `${resourceType} successfully created`;
+    notifications.show({
+      title: `${resourceType} Created!`,
+      message: message,
+      icon: <CircleCheck />,
+      color: 'green'
+    });
+    router.push(`/authoring/${resourceType}/${data.draftId}`);
+    ctx.draft.getDraftCounts.invalidate();
+  };
```
Comment (sarahmcdougall) [inline app/src/pages/[resourceType]/[id].tsx:98]: I noticed that the release notification gets called on the measure and its child artifacts upon release, which is pretty cool because that gives another indication that the child artifacts are being accounted for. When I drafted a measure, I believe I only got a notification of the measure being drafted, and not of the child library also being drafted. What are your thoughts in regard to adding a notification for when a child is drafted? Do you think that would take a lot of re-working and be out out scope? At first glance it seems like this might be more difficult to achieve as opposed to the release notifications because the draft notification is built into the `[id]` page and not a component.

app/src/server/trpc/routers/service.ts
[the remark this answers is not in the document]
sarahmcdougall: For now I think it would be most helpful to put logs in the `draftChildren` and `releaseChildren` functions specifically logging that a child artifact is being drafted/released, perhaps along with their id. I don't think we need to add super extensive logging since we do have logging for the service operations already - would be helpful to have a few additional logs indicating whether a resource is a child artifact or not so that there's a little more context to the logs.

sarahmcdougall (changes requested):

app/src/server/trpc/routers/service.ts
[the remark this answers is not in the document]
elsaperelli: This was the one thing I had a lot of trouble with adding and figured I could either continue working on it after I restructured stuff or even put it in a different task...I tried loggerLink and it looks like it just formatted it in the browser console...but we want logs in the console with the server logs, right?

app/src/pages/[resourceType]/[id].tsx
Comment (elsaperelli) [reply app/src/pages/[resourceType]/[id].tsx]: Thank you for commenting about this! Unfortunately, it was quite a bit of reworking but it was all very necessary changes that not only accomplish what you suggested here, but also make the code a lot cleaner.

app/src/pages/[resourceType]/[id].tsx
```diff
@@ -84,30 +83,83 @@ export default function ResourceIDPage({ jsonData }: InferGetServerSidePropsType
...
-        title: `Draft ${jsonData.resourceType} Created!`,
-        message: `Draft ${jsonData.resourceType}/${jsonData.id} successfully created`,
-        icon: <CircleCheck />,
-        color: 'green'
-      });
-      router.push(`/authoring/${jsonData.resourceType}/${data.draftId}`);
-      ctx.draft.getDraftCounts.invalidate();
+  const successNotification = (
+    resourceType: string,
+    createdFromArtifact: boolean,
```
Comment (sarahmcdougall) [inline app/src/pages/[resourceType]/[id].tsx:88]: I'm having a hard time thinking of a situation when `createdFromArtifact` will be `false` here since we are creating a draft of a specific artifact from this page. What are examples where `createdFromArtifact` will be false? Or is it only the case that this input will be false when we are on the authoring page (which I think has its own notification functions)?

app/src/pages/authoring/index.tsx
```diff
@@ -66,20 +86,39 @@ export default function AuthoringPage() {
...
-      errorNotification(e.message, false);
+    onError: (e, variables) => {
+      errorNotification(variables.resourceType, e.message, false, false);
     }
   });
 
-  const draftFromArtifactMutation = trpc.service.convertArtifactById.useMutation({
-    onSuccess: data => {
-      successNotification(data, true);
+  const draftChildMutation = trpc.service.draftChild.useMutation({
```
sarahmcdougall, line 98: Might be worth adding a comment to this function regarding the assumptions we make with failing child artifacts. This will help us in the future if we need to look back and see where the assumptions come into play with the actual code.

app/src/components/ReleaseModal.tsx
```diff
@@ -21,78 +20,114 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
...
+    },
+    onError: (e, variables) => {
+      console.error(e);
       notifications.show({
         title: `Release Failed!`,
-        message: `Attempt to release ${resourceType} failed with message: ${e.message}`,
+        message: `Attempt to release ${variables.resourceType} failed with message: ${e.message}`,
         icon: <AlertCircle />,
         color: 'red'
       });
```
sarahmcdougall, line 68: As mentioned in our earlier discussion, we could add more detail to this error message in regard to which child artifact failed so that the user can easily navigate to it and know that it is a child artifact of the parent

sarahmcdougall (changes requested):
Great job re-working our approach here for handling child artifacts! I appreciate all the thought that was put into the design for this, and I find the notifications to be especially helpful for detailing the process. I have a few smaller comments but overall the functionality works as expected!

app/src/server/trpc/routers/service.ts
```diff
@@ -1,9 +1,65 @@
...
+  url: string;
+  version: string;
+};
+
+/**
+ * Helper function that takes an array of related artifacts and recursively
+ * finds all of the child artifacts (related artifacts whose type is composed-of
+ * and has the isOwned extension)
+ */
+async function getChildren(relatedArtifacts: fhir4.RelatedArtifact[]) {
```
lmd59, line 20: I suspect that we want to keep this file focused on procedures if we can help it. Can we make a helper or utility file instead and put this function there?

app/src/server/trpc/routers/service.ts
```diff
@@ -62,13 +118,111 @@ export const serviceRouter = router({
...
+  // new release procedures (separate release of parent and release of child)
+  releaseParent: publicProcedure
+    .input(z.object({ resourceType: z.enum(['Measure', 'Library']), id: z.string(), version: z.string() }))
+    .mutation(async ({ input }) => {
+      // get the parent draft artifact by id
+      const draftRes = await getDraftById(input.id, input.resourceType);
+
+      // if a draft artifact of that resource type and id exists, then modify its content to be an active artifact
+      if (draftRes) {
+        draftRes.version = input.version;
```
lmd59, line 177: I think out of scope for this task but maybe worth thinking about: I don't think we allow editing on the version, so theoretically we could end up releasing with a version that already exists on the Measure repo, and the user has no way of changing it.
Example: 0.0.001 and 0.0.002 exist in repo. 0.0.001 is used to create a draft, which is auto assigned version 0.0.002. 0.0.002 draft is released and collides with existing 0.0.002 in the repo.

app/src/server/trpc/routers/service.ts
lmd59: Or actually I think maybe it makes sense to make this its own endpoint (query procedure) since it's doing MRS calls to get all of the information.

lmd59 (commented):
Have a couple of high level comments and agree with Sarah's. Mostly looks good! Well done reorganizing things!

app/src/util/types/fhir.ts
```diff
@@ -1,8 +1,10 @@
 export interface CQFMMeasure extends fhir4.Measure {
+  id: string;
```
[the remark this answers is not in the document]
elsaperelli: It's a bit confusing in the spec actually because I agree that `id` is not required for the CQFM profiles; however, it seems as though `id` is not just required for our implementation. There are multiple places in the spec that seem to insinuate that a server-defined `id` is required. There are some spots:
- _logical id_: https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#artifact-identity
- SHOULD params for requriements: https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#requirements
- SHOULD params for package: https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#package

Am I interpreting this correctly? Happy to add a comment if you think that would be helpful. If we remove it from the interface, the typing some places gets tricky because id will be `string | undefined` but we know it must be defined.

app/src/server/trpc/routers/service.ts
elsaperelli: I totally agree that this is something we will have to address. This is documented (or at least mostly) in [this task](https://jira.mitre.org/browse/TACOSTRAT-1386), so I will hold off on this functionality if you think that's okay.

lmd59: Cool! I just added a comment to the task that "We also need released artifacts to not conflict with drafts." since the task seems to just be enforcing unique urls across drafts.

app/src/server/trpc/routers/service.ts
lmd59: Oh totally- sorry! I didn't the repercussions of that second comment completely through. Ignore and go with helper or utility!

lmd59 (approved):

PR #72, opened 2023-12-29
Add support for directory upload

sarahmcdougall:
# Summary
Adds support for directory bundle upload via both the `db:loadBundle` npm script as well as a new batch script.

## New behavior
The user now has the ability to upload entire directories of bundles to the measure repository rather than needing to run a command on each individual JSON file.

The existing `db:loadBundle` script can be used for directory upload and supports nested directories (one level deep). The script detects whether the input path refers to a file or directory. The new bash script also supports nested directories and applies a transaction bundle upload via a curl command to each of the bundles contained in the directory.

## Code changes
* `service/scripts/dbSetup.ts` - update script to check whether the input path is a directory, and loop over all JSON files accordingly
* `service/directory-upload.sh` - new bash script that takes in a directory path and server url, and applies transaction bundle upload to each JSON bundle in the directory
* `service/README.md` - update documentation on usage of `db:loadBundle` script, details on how the bundles are uploaded (in regard to additions to the data with the isOwned extension, etc.), details on transaction bundle upload

# Testing guidance
`db:loadBundle` testing:
* Reset database, then run script with a file and ensure that functionality remains unchanged
* Reset database, then run script with a directory containing JSON bundles and ensure that the resources in the bundles are successfully uploaded to server
* Reset database, then run script with a directory containing sub-directories of JSON bundles and ensure that the script can traverse the nested directories. For testing, I use `ecqm-content-r4-2021/bundles/measure`

**Open question(s) regarding `db:loadBundle` implementation:**
* Should we throw an error if an artifact is an uploaded artifact is not in “active” status? I have noticed during previous testing with the measure repository that “draft” artifacts may be uploaded to the repository via this script. Would it be preferable to throw a warning about coercing the version to “active,” and then update the status to “active” within this script?

Bash script testing:
* Test the various script options to check that they are supported and function as intended. Try testing invalid options to check that they are properly handled.
* Example command to run script from the terminal: `./directory-upload.sh -s "http://localhost:3000/4_0_1" -d ./ecqm-content-r4-2021/bundles/measure/ColorectalCancerScreeningsFHIR`
* Check that the script can handle nested directories (can similarly use the ecqm-content repository to test this)

**Open question(s) regarding bash script/transaction bundle upload implementation:**
* Currently a 400 error is thrown in `BaseService.ts` if a bundle contains non-Measure/non-Library resources. Should we add documentation requiring that the bundles specified into the bash script only contain Measure and Library resources?
    * Can resources in a transaction bundle be “skipped” during upload? Could we skip non-Measure/non-Library resources, and return a 400 for those entries within the transaction-response bundle? Right now it seems like `BaseService.ts` does not do any error handling in the transaction-response.

Additional testing recommendations:

- Test using the connectathon measures at https://github.com/cqframework/ecqm-content-qicore-2024 i.e. `./directory-upload.sh -s "http://localhost:3000/4_0_1" -d ../../ecqm-content-qicore-2024/bundles/mat/`
- Test the operation outcome warning by posting a bundle that has a draft measure but no valuesets such as CMS130FHIR-v0.1.000-FHIR (colon cancer screening) from qicore-2024. The response should show an entry for the Measure with the warning operation outcome for coercion to active status.

### DIFF
--- a/service/directory-upload.sh
+++ b/service/directory-upload.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+usage="
+usage: $(basename "$0") command [-h] [-s] [-d] arguments...
+Uploads Measure and Library resources to the measure repository service.
+Options:
+    -h
+        Displays help menu.
+    -s [server baseUrl]
+        Specifies the base URL of the FHIR server to access.
+    -d
+        Provides directory path to parse for upload.
+"
+
+while getopts ':h:s:d:' option;
+do
+  case "$option" in
+    h)
+       echo -e "$usage"
+       exit 0
+       ;;
+    s)
+       server=$OPTARG
+       ;;
+    d)
+       directory_path=$OPTARG
+       ;;
+
+   \?) printf "illegal option: -%s\n" "$OPTARG" 1>&2
+       echo "$usage" 1>&2
+       exit 1
+       ;;
+    : )
+      echo "Invalid option: $OPTARG requires an argument" 1>&2
+      ;;
+  esac
+done
+
+if [[ $directory_path == "" ]] ; then
+  echo No directory path provided. Provide directory path via the '-d' flag.
+  exit 1
+fi
+
+if [[ $server == "" ]] ; then
+  echo No server URL provided. Provide server URL via the '-s' flag.
+  exit 1
+fi
+
+echo Using Server URL: $server and directory path: $directory_path
+
+upload_bundle() {
+  echo "Uploading resources for bundle $1"
+    # TODO: update the url or even make that an input...?
+    curl_command="curl -X POST -H 'Content-Type: application/json+fhir' -d @\"$1\" $server"
+    # execute the curl command
+    eval "$curl_command"
+
+    echo "Bundle successfully uploaded."
+    echo ""
+}
+
+# loop over FHIR bundles in specified directory
+for file_path in "$directory_path" ; do
+  if [[ -d $file_path ]] ; then
+    # recurse on directory
+    for f in $(find $file_path -name "*.json") ; do
+      upload_bundle $f
+    done
+
+  elif [[ -f $file_path ]] ; then
+    if [[ ${file_path: -5} == ".json" ]] ; then
+      upload_bundle $file_path
+    fi
+  fi
+done
--- a/service/directory-upload.sh
+++ b/service/directory-upload.sh
@@ -12,7 +12,7 @@ Options:
         Provides directory path to parse for upload.
 "
 
-while getopts ':h:s:d:' option;
+while getopts ':hs:d:' option;
 do
   case "$option" in
     h)
--- a/service/directory-upload.sh
+++ b/service/directory-upload.sh
@@ -8,8 +8,8 @@ Options:
         Displays help menu.
     -s [server baseUrl]
         Specifies the base URL of the FHIR server to access.
-    -d
-        Provides directory path to parse for upload.
+    -d [path]
+        Provides directory or file path to parse for upload.
 "
 
 while getopts ':hs:d:' option;
--- a/service/directory-upload.sh
+++ b/service/directory-upload.sh
@@ -54,7 +54,7 @@ upload_bundle() {
     # execute the curl command
     eval "$curl_command"
 
-    echo "Bundle successfully uploaded."
+    echo "Finished bundle upload."
     echo ""
 }
 
--- a/service/directory-upload.sh
+++ b/service/directory-upload.sh
@@ -50,8 +50,7 @@ echo Using Server URL: $server and directory path: $directory_path
 
 upload_bundle() {
   echo "Uploading resources for bundle $1"
-    # TODO: update the url or even make that an input...?
-    curl_command="curl -X POST -H 'Content-Type: application/json+fhir' -d @\"$1\" $server"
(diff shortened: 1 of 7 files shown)


service/directory-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+
+usage="
+usage: $(basename "$0") command [-h] [-s] [-d] arguments...
+Uploads Measure and Library resources to the measure repository service.
+Options:
+    -h
+        Displays help menu.
+    -s [server baseUrl]
+        Specifies the base URL of the FHIR server to access.
+    -d
```
lmd59, line 11:
```suggestion
    -d [path]
```

service/directory-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+    d)
+       directory_path=$OPTARG
+       ;;
+
+   \?) printf "illegal option: -%s\n" "$OPTARG" 1>&2
+       echo "$usage" 1>&2
+       exit 1
+       ;;
+    : )
+      echo "Invalid option: $OPTARG requires an argument" 1>&2
```
lmd59, line 34: I hit this option only with a -h (no argument), which I'm assuming should just show the usage.

service/directory-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+usage="
+usage: $(basename "$0") command [-h] [-s] [-d] arguments...
+Uploads Measure and Library resources to the measure repository service.
+Options:
+    -h
+        Displays help menu.
+    -s [server baseUrl]
+        Specifies the base URL of the FHIR server to access.
+    -d
+        Provides directory path to parse for upload.
```
lmd59, line 12: I believe we can also use a file path
```suggestion
        Provides directory or file path to parse for upload.
```

lmd59 (commented):
A couple of comments, and looking into answers to existing questions

lmd59:
Decisions on questions:
1. We've decided to coerce the status to active where needed.
2. We're okay with throwing an error if valuesets are included in the bundle. The server only accepts Library or Measure and no other resource types.

service/directory-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+if [[ $server == "" ]] ; then
+  echo No server URL provided. Provide server URL via the '-s' flag.
+  exit 1
+fi
+
+echo Using Server URL: $server and directory path: $directory_path
+
+upload_bundle() {
+  echo "Uploading resources for bundle $1"
+    curl_command="curl -X POST -H 'Content-Type: application/json+fhir' -d @\"$1\" $server -o /dev/null"
```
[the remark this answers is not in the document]
elsaperelli: Something I noticed that I believe solves this problem are the `logger.debug` statements in `baseUtils.ts`...now that we have a script that does a ton of requests, perhaps we should modify these? When I comment them out, I no longer get my error and I can better see the requests that are being made.

service/src/services/BaseService.ts
```diff
@@ -63,6 +63,21 @@ async function uploadResourcesFromBundle(entries: DetailedEntry[]) {
  */
 async function insertBundleResources(entry: DetailedEntry) {
   if (entry.resource?.resourceType === 'Library' || entry.resource?.resourceType === 'Measure') {
+    if (entry.resource.status != 'active') {
```
elsaperelli, line 66: Do we want this for any Measure or Library we insert in the database? Not just the ones from a bundle?

elsaperelli (changes requested):
Great stuff Sarah and Lauren! I think it's pretty awesome that we can use the bash script to upload the bundles in the `ecqm-content-qicore-2024` repository. I ran into a couple things but I think I have mostly figured them out (explained in the comments).

service/src/services/BaseService.ts
lmd59: I think that Hoss confirmed yes on that at dev hour.

service/directory-upload.sh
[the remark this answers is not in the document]
lmd59: With an ENOBUFS error, it seems like the buffer has grown too large. Eliminating some of the logging then may have limited the buffer growth and allowed this to be successful. This is something where handling may differ from system to system because of differences in resources available (memory) which could have been the cause of you seeing the error, but me being unable to reproduce it. We may be able to do some reorganization in the code, queueing, or maybe just eliminating extraneous logging to help prevent it coming up for you (and potentially on other systems).

service/directory-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+    # execute the curl command
+    eval "$curl_command"
+
+    echo "Bundle successfully uploaded."
```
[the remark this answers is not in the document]
lmd59: I tweaked this a bit just to indicate that it was finished (not necessarily successful).

service/src/services/BaseService.ts
lmd59: I added code to coerce to active for a regular create/update (non-bundle). Tried a couple of different approaches to get a similar OperationOutcome, but because we're not creating responses in the bundle it's a bit trickier, and my approaches messed up other things that were a part of the base functionality. I have those attempts locally stashed, so happy to discuss further, or just let it be with the logged warning rather than the operation outcome.

sarahmcdougall:
Just got caught up on recent progress with this PR. Thank you for the insightful comments and changes! I agree with all changes made thus far.

elsaperelli (approved):
Lgtm!!

lmd59 (approved):

PR #75, opened 2024-01-10
Added _summary search param for artifact count optimization

elsaperelli:
# Summary
This PR adds the `_summary` search parameter (specified in the HL7 FHIR docs [here](https://www.hl7.org/fhir/search.html#_summary)) to the Library and Measure search capabilities. Adding this allows our `getArtifactCounts` trpc procedure on the frontend to get the number of artifacts without sending a request that gets all of the artifacts.

## New behavior
Before, it took awhile for the frontend application to load the number of Measure and Library artifacts. This was due to how our `getArtifactCounts` trpc procedure was written- a GET request was sent to get all of the Measures and Libraries and then `.total` on the searchset bundle was returned. Now, `getArtifactCounts` sends requests to `/Measure?_summary=count` and `/Library?_summary=count` so that searchsets excluding all of the entries are returned. 

## Code changes
- `app/src/server/trpc/routers/service.ts` - changed the request to include the `_summary=count` parameter 
- `service/README.md` - added to the list of supported search params
- `service/src/db/dbOperations.ts` - added function `findResourceCountWithQuery` to get the number of documents (artifacts) a query returns
- `service/src/requestSchemas.ts` - added `_summary` as a CoreSearchArg
- `LibraryService.ts`/`MeasureService.ts` - added to `search` to return a summary searchset bundle if the `_summary=count` parameter is included in the query
- `service/src/util/bundleUtils.ts` - added `createSummarySearchsetBundle` function to create a searchset bundle with a specified count and without the `entry` array

# Testing guidance
- Upload a bunch of measure bundles (`npm run db:loadBundle <path to ecqm-content-qicore-2024/bundles/measure>` as an example in the `service` directory)
- Spin up server and frontend in the root directory: `npm run start:all`
- Navigate to the app in the browser. See how long it takes for the Measure and Library counts in the side panel to load, hopefully it is noticeably faster than the main branch.

- `npm run check:all` in root directory 

- Test out the `_summary` search parameter in Insomnia (or some other API client). Make sure that `_summary=count` works properly for both Measures and Libraries. Also check that parameter being used with other search parameters supported by the server.

# Questions
1. `_summary` is not listed in the [Measure Repository Service Docs](http://hl7.org/fhir/us/cqfmeasures/measure-repository-service.html#search) as a parameter. Is it okay that I do this then? I wasn't sure what an alternative solution would be for this issue since this seemed more correct than creating my own endpoint.

### DIFF
--- a/service/src/services/LibraryService.ts
+++ b/service/src/services/LibraryService.ts
@@ -3,12 +3,13 @@ import {
   createResource,
   findDataRequirementsWithQuery,
   findResourceById,
+  findResourceCountWithQuery,
   findResourcesWithQuery,
   updateResource
 } from '../db/dbOperations';
 import { LibrarySearchArgs, LibraryDataRequirementsArgs, PackageArgs, parseRequestSchema } from '../requestSchemas';
 import { Service } from '../types/service';
-import { createLibraryPackageBundle, createSearchsetBundle } from '../util/bundleUtils';
+import { createLibraryPackageBundle, createSearchsetBundle, createSummarySearchsetBundle } from '../util/bundleUtils';
 import { BadRequestError, ResourceNotFoundError } from '../util/errorUtils';
 import { getMongoQueryFromRequest } from '../util/queryUtils';
 import {
@@ -39,8 +40,16 @@ export class LibraryService implements Service<fhir4.Library> {
     logger.debug(`Request Query: ${JSON.stringify(query, null, 2)}`);
     const parsedQuery = parseRequestSchema(query, LibrarySearchArgs);
     const mongoQuery = getMongoQueryFromRequest(parsedQuery);
-    const entries = await findResourcesWithQuery<fhir4.Library>(mongoQuery, 'Library');
-    return createSearchsetBundle(entries);
+
+    // if the _summary parameter with a value of count is included, then
+    // return a searchset bundle that excludes the entries
+    if (parsedQuery._summary && parsedQuery._summary === 'count') {
+      const count = await findResourceCountWithQuery(mongoQuery, 'Library');
+      return createSummarySearchsetBundle<fhir4.Library>(count);
+    } else {
+      const entries = await findResourcesWithQuery<fhir4.Library>(mongoQuery, 'Library');
+      return createSearchsetBundle(entries);
+    }
   }
 
   /**
--- a/service/src/services/MeasureService.ts
+++ b/service/src/services/MeasureService.ts
@@ -3,11 +3,12 @@ import {
   createResource,
   findDataRequirementsWithQuery,
   findResourceById,
+  findResourceCountWithQuery,
   findResourcesWithQuery,
   updateResource
 } from '../db/dbOperations';
 import { Service } from '../types/service';
-import { createMeasurePackageBundle, createSearchsetBundle } from '../util/bundleUtils';
+import { createMeasurePackageBundle, createSearchsetBundle, createSummarySearchsetBundle } from '../util/bundleUtils';
 import { BadRequestError, ResourceNotFoundError } from '../util/errorUtils';
 import { getMongoQueryFromRequest } from '../util/queryUtils';
 import {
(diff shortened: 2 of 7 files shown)


service/src/services/LibraryService.ts
```diff
@@ -39,8 +40,16 @@ export class LibraryService implements Service<fhir4.Library> {
...
     const mongoQuery = getMongoQueryFromRequest(parsedQuery);
-    const entries = await findResourcesWithQuery<fhir4.Library>(mongoQuery, 'Library');
-    return createSearchsetBundle(entries);
+
+    // if the _summary parameter with a value of count is included, then
+    // return a searchset bundle that excludes the entries
+    if (parsedQuery._summary && parsedQuery._summary === 'count') {
+      const count = await findResourceCountWithQuery(mongoQuery, 'Library');
+      return createSummarySearchsetBundle<fhir4.Library>(count);
+    } else {
```
lmd59, line 49: It looks like we default to a normal searchset bundle if the query includes the `_summary` parameter but the value isn't `count`. Do we want to throw an unsupported error for that case instead?

service/src/services/MeasureService.ts
```diff
@@ -40,8 +41,16 @@ export class MeasureService implements Service<fhir4.Measure> {
...
     const mongoQuery = getMongoQueryFromRequest(parsedQuery);
-    const entries = await findResourcesWithQuery<fhir4.Measure>(mongoQuery, 'Measure');
-    return createSearchsetBundle(entries);
+
+    // if the _summary parameter with a value of count is included, then
+    // return a searchset bundle that excludes the entries
+    if (parsedQuery._summary && parsedQuery._summary === 'count') {
+      const count = await findResourceCountWithQuery(mongoQuery, 'Measure');
+      return createSummarySearchsetBundle<fhir4.Measure>(count);
+    } else {
```
lmd59, line 50: It looks like we default to a normal searchset bundle if the query includes the `_summary` parameter but the value isn't `count`. Do we want to throw an unsupported error for that case instead?

service/src/services/LibraryService.ts
lmd59: Example: _summary=data

lmd59 (commented):
Nice significantly faster load! I still notice that the load is very slow for viewing the full Library list (i.e. `/Library`, up to 10s for me), but I think that's out of scope for this ticket. Let me know if you see the same behavior, and we can think about making a new ticket for that.

I made one small suggestion for an error state, and otherwise, I think this looks good.

In response to question: I think that this is at least the right choice for usability for the connectathon, so I would advocate for approval at this point either way. But because `summary` is not listed under the MAY support search parameters, I would want to check in with someone with a little bit more familiarity with the restrictions of the spec. I.e. is there any feedback we should be giving to the spec to include `_summary` in the MAY support documentation?

service/src/services/LibraryService.ts
elsaperelli: Good call!

elsaperelli:
> Nice significantly faster load! I still notice that the load is very slow for viewing the full Library list (i.e. `/Library`, up to 10s for me), but I think that's out of scope for this ticket. Let me know if you see the same behavior, and we can think about making a new ticket for that.
> 
> I made one small suggestion for an error state, and otherwise, I think this looks good.
> 
> In response to question: I think that this is at least the right choice for usability for the connectathon, so I would advocate for approval at this point either way. But because `summary` is not listed under the MAY support search parameters, I would want to check in with someone with a little bit more familiarity with the restrictions of the spec. I.e. is there any feedback we should be giving to the spec to include `_summary` in the MAY support documentation?

Thanks for the feedback Lauren!! I agree, the `/Library` does take awhile, but that this is out of scope for this task. I also think with may be a bit more challenging than what I did here since we use fields in the Library list (`id`, etc.) that we get some the Library json itself. We can make a task though!

I think I came up with a nifty solution (shoutout zod) for your error handling suggestion, but let me know if there is something else that makes more sense.

In terms of reaching out to someone about `_summary`, did you mean before the Connectathon? If not, I can just add it to the running list of MRS questions we have.

PR #76, opened 2024-01-12
Add isOwned extension functionality to txn bundle upload

elsaperelli:
# Summary
Previously, we had only implemented the `isOwned` child artifact functionality when bundles were uploaded using the `dbSetup.ts` script (`npm run db:loadBundle`). Now that we have the `./directory-upload.sh` script, want want to make sure that the `isOwned` extension gets added when transaction bundles are uploaded.

## New behavior
The same thing that is done for adding the `isOwned` extension on the `relatedArtifacts` of Measures is now done for transaction bundle uploads as well as the `dbSetup.ts` script.

Note: there is more code that we can probably consolidate/make more reusable but I tried to keep it to the scope of the `isOwned` extension so that we can have this functionality for the Connectathon.

## Code changes
- `dbSetup.ts` - pulled out `isOwned` extension code and replaced it with a function call (`addIsOwnedExtension`)
- `BaseService.ts` - added `isOwned` functionality to the transaction bundle upload through a call to `addIsOwnedExtension`
- `baseUtils.ts` - added `isOwned` extension code to reusable function `addIsOwnedExtension`

# Testing guidance
- `npm run check:all`
- Make sure child artifacts functionality works in the following scenarios (see #69 PR for child artifact testing guidance):
- Uploading a measure bundle through the `dbSetup.ts` script (`npm run db:loadBundle`)
- Uploading a directory of measure bundles (uses transaction bundle upload) through `./directory-upload.sh` script
- Send a `POST` request to the server where the body contains a transaction bundle

### DIFF
--- a/service/src/util/baseUtils.ts
+++ b/service/src/util/baseUtils.ts
@@ -1,10 +1,10 @@
 import { v4 as uuidv4 } from 'uuid';
 import { loggers } from '@projecttacoma/node-fhir-server-core';
-import { OperationOutcome } from 'fhir/r4';
+import { FhirResource, OperationOutcome } from 'fhir/r4';
 
 const logger = loggers.get('default');
 
-export type DetailedEntry = fhir4.BundleEntry & {
+export type DetailedEntry = fhir4.BundleEntry<FhirResource> & {
   isPost: boolean;
   oldId?: string;
   newId: string;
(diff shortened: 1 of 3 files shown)


service/src/util/baseUtils.ts
```diff
@@ -14,6 +14,56 @@ export type DetailedEntry = fhir4.BundleEntry & {
   outcome?: OperationOutcome;
 };
 
+/**
+ * Add isOwned extension to the main Library reference on a Measure's relatedArtifacts
```
lmd59, line 18:
```suggestion
 * Checks entry for Measure type with library and adds isOwned extension to the main Library reference on a Measure's relatedArtifacts
```
Thought it might be helpful to note that this can be applied to any entry because it can do the checking whether the reference should be added. Feel free to change wording though!

lmd59 (approved):
One suggestion for a comment update, but otherwise looks good and works good!

PR #78, opened 2024-01-22
Enforce unique combined url and version

lmd59:
# Summary
Enforced unique combination of url and version for both draft artifacts and measure repository artifacts. Updates some error handling to show useful error information to the user.

## New behavior
If a user attempts to create a resource (new or update or through operations or http PUT/POST) that has a combined url/version that conflicts with an existing combined url/version, they will receive an error indicating that the action failed because of a duplicate key (and will show the duplicate keys in question).

## Code changes

dbSetup.ts - creates a unique url/version index for the measure-repository collections
mongodb.ts - creates a unique url/version index for the draft-repository collections
service.ts - passes an error through if a 500 response is received from the server
ReleaseModal.tsx - on release, prioritizes check for error status (non 201) before checking for location and shows error message if any is passed through

# Testing guidance

`npm run check:all`
Recommend dropping/resetting both databases before testing so that indexes are created fresh.

For the following tests, an error with something like "E11000 duplicate key error collection: draft-repository.Measure index: url_1_version_1 dup key: { url: "www.test.com", version: "0.0.1" }" or similar is expected (but with any additional context and the specific keys that are of issue in each case).

- Draft artifacts: Test draft artifact conflicts by creating a new draft measure and giving it a url, such as 'www.test.com'. Create another measure and give it the same url. You should receive an error notification upon hitting the "Update" button.
- Artifact release: Create a draft artifact with the same url as an active artifact. You can do this by releasing an existing active Measure (i.e. load the Colonoscopy Measure, view, and select "Create Draft of Measure"). Manually update the version of the draft artifact to match the active artifact (I used Robo 3T to update this directly in the database). Release the draft measure. You should receive an error notification upon hitting the "Release" button.
- Artifact child release: Similar to the above, but instead, change the version of the draft child library to match the active child library, and return the parent draft measure version to an incremented version that doesn't match the active measure. Release the draft measure. You should receive a success notification for the measure and an error notification for the child library.
- Artifact draft: Use the "Create Draft of Measure" (or Library) button twice in a row on the same artifact. The second time should show an error notification.
- Service interaction: You can also test the server unique keys with direct insomnia POST (or PUT). For example, copy the content (or at least use the same url and version) of an existing active Measure. Put that content in a POST to http://localhost:3000/4_0_1/Measure. The server should return a 500 error with an OperationOutcome indicating the duplicate key.
- Play around with other server interactions or any other user actions that can trigger an artifact key conflict.

### DIFF
--- a/service/src/services/MeasureService.ts
+++ b/service/src/services/MeasureService.ts
@@ -183,6 +183,7 @@ export class MeasureService implements Service<fhir4.Measure> {
     // add the data requirements query params to the data requirements Library resource and add to the Library collection
     const results = { ...dataRequirements.results } as FhirLibraryWithDR;
     results['_dataRequirements'] = dataReqsQuery;
+    results.url = `data-requirements/${dataRequirements.results.id}`;
     createResource(results, 'Library');
 
     logger.info('Successfully generated $data-requirements report');
--- a/service/src/services/MeasureService.ts
+++ b/service/src/services/MeasureService.ts
@@ -183,7 +183,7 @@ export class MeasureService implements Service<fhir4.Measure> {
     // add the data requirements query params to the data requirements Library resource and add to the Library collection
     const results = { ...dataRequirements.results } as FhirLibraryWithDR;
     results['_dataRequirements'] = dataReqsQuery;
-    results.url = `data-requirements/${dataRequirements.results.id}`;
+    results.url = `Library/${dataRequirements.results.id}`;
     createResource(results, 'Library');
 
     logger.info('Successfully generated $data-requirements report');
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -36,7 +36,7 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
         console.error(data.status);
         notifications.show({
           title: `Release Failed!`,
-          message: `Server unable to process request`,
+          message: `Server unable to process request. ${data.error ?? ''}`,
           icon: <AlertCircle />,
           color: 'red'
         });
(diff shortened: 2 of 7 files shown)


elsaperelli (commented):
I am still having a hard time exactly pin-pointing this error, but I think the caching data-requirements code is conflicting with the new unique version and url code somehow. When I navigate to a Measure or a Library multiple times, I get a duplicate key error that causes the app to crash:

image

Let me know if you can replicate this.

lmd59:
> I am still having a hard time exactly pin-pointing this error, but I think the caching data-requirements code is conflicting with the new unique version and url code somehow. When I navigate to a Measure or a Library multiple times, I get a duplicate key error that causes the app to crash:
> 
> image
> 
> Let me know if you can replicate this.

Adding a url for the cached data-requirements library should fix this. Let me know if you think this url makes sense, or if we should do a different url, or we could even just strip it off again when we pull it from cache if a url doesn't makes sense at all

service/src/services/MeasureService.ts
```diff
@@ -183,6 +183,7 @@ export class MeasureService implements Service<fhir4.Measure> {
     // add the data requirements query params to the data requirements Library resource and add to the Library collection
     const results = { ...dataRequirements.results } as FhirLibraryWithDR;
     results['_dataRequirements'] = dataReqsQuery;
+    results.url = `Library/${dataRequirements.results.id}`;
```
elsaperelli, line 186: This is nitpicky but good to make sure we have consistent now I think...

When I navigate to the data requirements tab of a Measure/Library for the first time (aka it is not yet cached), then I scroll to the bottom and I see "id" that was inserted there (line 181) but I do not see the url. If I then navigate away and then back to that artifact's data requirements tab, then the url is there. In my opinion we should do one of two things:

1. Make the url "hidden", we do this with _dataRequirements by adding a projection of 0 in the `dbOperations.ts` file. I personally am leaning towards this one since the url is "made up" and I think we want to display the data requirements as a Library that follows the spec as much as possible.
2. If we want to display the url, then I would make it so that you can see the url when you first navigate to the data requirements tab. You can do this by doing `dataRequirements.results['url']...` since that is what gets returned.

lmd59: I think that this url may be accurate to the spec because it's the relative location of the data-requirements library on the server, but I'm not 100% sure. I went with option 1 for now, but if there's a reason that we think that the url might be useful to a user at some point, we can change things around to use option 2.

elsaperelli (approved):
Good stuff!!

PR #80, opened 2024-02-06
Add version to the artifact name/id in the dropdown

elsaperelli:
# Summary
Adds version to the label of an artifact so it's not just the name or id. This prevents multiple artifacts of the same name but with different versions appearing in the dropdown for drafting measures:

image

## New behavior
None

## Code changes
- `service.ts` - add version to the artifact label

# Testing guidance
- `npm run check:all`
- `npm run start:all`
- Make sure everything looks good!

### DIFF
--- a/app/src/server/trpc/routers/service.ts
+++ b/app/src/server/trpc/routers/service.ts
@@ -37,7 +37,11 @@ export const serviceRouter = router({
         resArtifacts => resArtifacts.json() as Promise<fhir4.Bundle<FhirArtifact>>
       );
       const artifactList = artifactBundle.entry?.map(entry => ({
-        label: entry.resource?.name || entry.resource?.id || '',
+        label:
+          entry.resource?.name?.concat(`|${entry.resource?.version}`) ||
+          entry.resource?.name ||
+          entry.resource?.id ||
+          '',
         value: entry.resource?.id || `${entry.resource?.resourceType}` || '',
         disabled: !!entry.resource?.extension?.find(
           ext => ext.url === 'http://hl7.org/fhir/StructureDefinition/artifact-isOwned' && ext.valueBoolean === true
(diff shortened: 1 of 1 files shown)


app/src/server/trpc/routers/service.ts
```diff
@@ -37,7 +37,11 @@ export const serviceRouter = router({
         resArtifacts => resArtifacts.json() as Promise<fhir4.Bundle<FhirArtifact>>
       );
       const artifactList = artifactBundle.entry?.map(entry => ({
-        label: entry.resource?.name || entry.resource?.id || '',
+        label:
+          entry.resource?.name?.concat(`|${entry.resource?.version}`) ||
+          entry.resource?.name ||
```
lmd59, line 42: I don't think we need this `entry.resource?.name ||` line because the first line is already checking whether there's a name field.

app/src/server/trpc/routers/service.ts
```diff
@@ -37,7 +37,11 @@ export const serviceRouter = router({
         resArtifacts => resArtifacts.json() as Promise<fhir4.Bundle<FhirArtifact>>
       );
       const artifactList = artifactBundle.entry?.map(entry => ({
-        label: entry.resource?.name || entry.resource?.id || '',
+        label:
+          entry.resource?.name?.concat(`|${entry.resource?.version}`) ||
```
lmd59, line 41:
```suggestion
          entry.resource?.name?.concat(`|${entry.resource.version}`) ||
```
don't need the second resource `?` because it's already been checked in this line

app/src/server/trpc/routers/service.ts
```diff
@@ -37,7 +37,11 @@ export const serviceRouter = router({
...
+          entry.resource?.name ||
+          entry.resource?.id ||
```
lmd59, line 43: Do we also want to concat the version on the id if that's used instead?

app/src/server/trpc/routers/service.ts
elsaperelli: Good point! I thought we would want to display the name if there wasn't a name AND a version but I don't think that'll ever be the case (we make sure there is a version on upload).

lmd59 (approved):
Nice!

PR #82, opened 2024-02-09
Batch release

lmd59:
# Summary
Child resources are batched with their parent for release.

## New behavior
The front end acts the same with some minor tweaks to notifications. Under the hood, the parent and child artifacts are released using a transaction bundle in order to ensure they are batched.

## Code changes

- ReleaseModal - releaseChild mutation is removed, transaction bundles response expected to be a 200, a notification is shown and a resource deleted for every released resource in the `deletable` list
- service.ts - construct a transaction bundle, and create a toDelete structure with the parent data, then iterate through the children to add to the transaction bundle entries and the toDelete structure. Send transaction bundle and use response to send new location and deletable resource information to the ReleaseModal

# Testing guidance

- `npm run check:all`
- in /service `npm run db:reset`
- in /service `npm run db:loadBundle {your favorite bundle here}` (I used ColorectalCancerScreenings)
- `npm run start:all`
- use the measure to create a draft (should also draft main library)
- release the draft measure (should also release the draft library, delete both, and redirect to the new released measure)

### DIFF
--- a/app/src/components/ReleaseModal.tsx
+++ b/app/src/components/ReleaseModal.tsx
@@ -30,47 +30,9 @@ export default function ReleaseModal({ open = true, onClose, id, resourceType }:
     }
   });
 
-  const releaseChildMutation = trpc.service.releaseChild.useMutation({
-    onSuccess: (data, variables) => {
-      if (data.status !== 201) {
-        console.error(data.status);
-        notifications.show({
-          title: `Release Failed!`,
-          message: `Server unable to process request. ${data.error ?? ''}`,
-          icon: <AlertCircle />,
-          color: 'red'
-        });
-      } else {
-        notifications.show({
-          title: `Draft ${variables.resourceType} released!`,
-          message: `Child draft ${variables.resourceType}/${data.id} successfully released to the Publishable Measure Repository!`,
-          icon: <CircleCheck />,
-          color: 'green'
-        });
-        ctx.draft.getDraftCounts.invalidate();
-        ctx.draft.getDrafts.invalidate();
-
-        // delete draft child artifact from the draft database now that it has been released
-        deleteMutation.mutate({
-          resourceType: variables.resourceType,
-          id: data.id
-        });
-      }
-    },
-    onError: (e, variables) => {
-      console.error(e);
-      notifications.show({
-        title: `Release Failed!`,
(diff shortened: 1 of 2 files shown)


elsaperelli (changes requested):
Clean code! This makes so much more sense- good job!

I did find two use cases that have some behavior that I am either not sure we want or maybe we can improve on?

1. If I have a Measure and its associated child Library artifact in draft and I edit the Library version in studio3T to be the same as the same one already in the publishable measure repository (I did this to hit the error handling), then when I go to release the Measure artifact and click "Release" in the modal, nothing happens. Perhaps we want to show a notification here or something? I know this is kind of a rare use case but at first I thought the app crashed or something so it was slightly misleading.

2. This use case is a little more challenging on how we want to handle it. If I have a Measure and its child Library artifact in draft, I can delete the Library (which actually we will need to change that in the future so it follows the child/parent artifact rules). If I then go to release the Measure, the modal stops and nothing happens (like the scenario above). 

After putting this all into writing, I don't think we need to worry about # 2 because we have to change how delete works with child artifacts anyways...but I do think we should have some sort of UI representation of batch release failure. If you agree, we can talk about what they may look like!

lmd59:
> Clean code! This makes so much more sense- good job!
> 
> I did find two use cases that have some behavior that I am either not sure we want or maybe we can improve on?
> 
> 1. If I have a Measure and its associated child Library artifact in draft and I edit the Library version in studio3T to be the same as the same one already in the publishable measure repository (I did this to hit the error handling), then when I go to release the Measure artifact and click "Release" in the modal, nothing happens. Perhaps we want to show a notification here or something? I know this is kind of a rare use case but at first I thought the app crashed or something so it was slightly misleading.
> 2. This use case is a little more challenging on how we want to handle it. If I have a Measure and its child Library artifact in draft, I can delete the Library (which actually we will need to change that in the future so it follows the child/parent artifact rules). If I then go to release the Measure, the modal stops and nothing happens (like the scenario above).
> 
> After putting this all into writing, I don't think we need to worry about # 2 because we have to change how delete works with child artifacts anyways...but I do think we should have some sort of UI representation of batch release failure. If you agree, we can talk about what they may look like!

Addressed 1 in the latest commit, and thanks for making a task for 2!

elsaperelli (approved):
🥳

PR #86, opened 2024-02-28
Version handling x.xx.xxx and small changes to increment logic

lmd59:
# Summary
Handle versioning x.xx.xxx for artifacts

## New behavior
x.xx.xxx is handled similar to x.x.xxx version. where both the patch version and the minor version are padded with 0s as needed, and next up (more major) version place is incremented if the more minor version exceeds its existing bounds. Additionally, these version formats are prioritized over semantic versioning, and there's a small update to address a bug in padding with zeroes for patch versions 10-99. 

## Code changes
In modifyResourceFields, `checkVersionFormat` allows for a two digit minor version. `modifyResourceToDraft` prioritizes MADiE format over semver. `incrementArtifactVersion` increments x.xx.xxx format and pads appropriately.

# Testing guidance
Load an artifact into the measure repository service and play with the version (I used robo 3T to edit it easily), then draft the artifact to cause version increments. Make sure version increments appropriately. Some suggested version cases:

- 1.1.1 (semantic)
- 1.1.001
- 1.01.001
- 1.99.001
- 1.99.999
- 1.1.999
- 1.01.999
- 1.11.999
- 1.1.009

### DIFF
--- a/app/src/util/modifyResourceFields.ts
+++ b/app/src/util/modifyResourceFields.ts
@@ -14,18 +14,18 @@ export async function modifyResourceToDraft(artifact: FhirArtifact) {
 
   // initial version coercion and increment
   // we can only increment artifacts whose versions are either semantic, can be coerced
-  // to semantic, or are in x.x.xxx format. Every other kind of version will become 0.0.1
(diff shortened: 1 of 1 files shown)


elsaperelli (approved):
Yay versioning!

PR #96, opened 2024-05-02
Add comment about changing hoster.docker.internal for linux systems

elsaperelli:
# Summary
This PR adds a comment to the `docker-compose.example.yml` file to reflect the change that we had to make in the `docker-compose.yml` in abacus-dev in order to get replicaSet functionality to work on a linux platform.

## New behavior
- n/a

## Code changes
- `docker-compose.example.yml` - adds comment, looks like prettier was applied too

# Testing guidance
- Make sure abacus-dev.mitre.org/mrs works as intended 
- Make sure this setup makes sense for now. Note that Lauren made a backlog task for looking into this further after our discussion.

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: "3"
 
 services:
   measure-service:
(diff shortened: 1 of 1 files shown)


lmd59 (approved):
Works for me!

PR #97, opened 2024-05-09
Update dates according to review spec

lmd59:
# Summary

Update review operation for draft artifacts (other statuses are still being check on) according to the [latest spec](https://github.com/HL7/crmi-ig/blob/master/input/fsh/operation-definitions/review-operation.fsh):

> The review operation supports applying an review to an existing artifact, regardless of status. The operation sets the _date_ and _lastReviewDate_ elements of the reviewed artifact, and is otherwise only allowed to create ArtifactAssessment (Basic or cqf-artifactComment extension in R4) resources in the repository.

Because we are working with R4, this means using artifactComments (we already are), but we can still update the logic to set the date and lastReviewDate elements of the reviewed artifact.

## New behavior
Creating a new review comment sets the date and lastReviewDate of the artifact.

## Code changes

- Updates to the README description of the review operation
- Changes to app/src/pages/review/[resourceType]/[id].tsx to allow the creation of a comment to update the date and lastReviewDate

# Testing guidance

- 'npm run check:all`
- `npm run start:all`
- Create a draft artifact if necessary
- Add a review comment to the draft artifact
- View the draft artifact json (or in the database with 3T or similar) and check that the date and lastReviewDate have been updated on the artifact


app/src/pages/review/[resourceType]/[id].tsx
```diff
@@ -100,6 +102,7 @@ export default function CommentPage() {
   function parseUpdate(comment: string, type: string, userName: string, dateSelected: boolean) {
     const additions: DraftArtifactUpdates = {};
     const deletions: DraftArtifactUpdates = {};
+    const currentDate = new Date().toISOString();
```
Comment (elsaperelli) [inline app/src/pages/review/[resourceType]/[id].tsx:105]: This is what we had before so not a big deal at all since I am just noticing, but is this the time that we want?

elsaperelli (changes requested):
Works and looks great! Thank you for implementing this ahead of the Connectathon!

Just two small comments.

elsaperelli (approved):
Lgtm!

PR #98, opened 2024-05-09
Pull out date change code

lmd59:
# Summary
Remove functionality that allows the user to change the date format on review comments.

## New behavior
Review comment date is not clickable and stays in the same format. This feature does not seem to have a lot of utility. Removing it fixes an issue with date displays where a review comment card may show the date of another review comment rather than its own. To reproduce this issue on the main branch:

- Add multiple review comments with dates to a draft artifact, taking note of their correct timestamps
- Add a review comment on the add tab
- Immediately switch over to the view tab
- Some review comments will have incorrect timestamps
- Timestamps are fixed with reload

## Code changes
Remove hover, color change, click response, and date formatter, replaced with simple date display.

# Testing guidance

- 'npm run check:all`
- Run with `npm run start:all`
- Create a draft artifact if necessary
- Select review option for draft artifact
- Try to reproduce issue described above (issue should be fixed, and accurate timestamps should be simply displayed)

### DIFF
--- a/app/src/components/ArtifactComments.tsx
+++ b/app/src/components/ArtifactComments.tsx
@@ -1,7 +1,5 @@
 import { Avatar, createStyles, Group, Paper, rem, Text, TypographyStylesProvider } from '@mantine/core';
 import parse from 'html-react-parser';
-import { useHover } from '@mantine/hooks';
-import { useEffect, useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


elsaperelli (approved):
Agreed that we no longer need this functionality. Lgtm!

PR #103, opened 2024-07-03
Expose  operation

elsaperelli:
# Summary
This PR exposes the $draft operation as an endpoint on the server. This functionality is defined in the CRMI IG [here](https://build.fhir.org/ig/HL7/crmi-ig/OperationDefinition-crmi-draft.html). 

## New behavior
The user can now create a draft version of a Measure or Library artifact and any resources it is composed of by sending a GET or POST request to `Measure/$draft`, `Measure/:id/$draft`, `Library/$draft` and `Library/:id/$draft`. Since this is only an operation supported in an Authoring Artifact Repository, the `AUTHORING` environment variable must be set to true.

## Code changes
- `service/env.test` / `service/jest.mongodb.config.js` / `service/package.json` - had to make some changes with the jest configuration since we need a replicaSet in order to test the batch draft functionality (was previously in the `/app` directory and there are no unit tests there) Source: https://github.com/shelfio/jest-mongodb?tab=readme-ov-file#2-create-jest-mongodb-configjs
- `service/src/config/serverConfig.ts` - added `$draft` endpoints to server configuration
- `service/src/db/dbOperations` - added `findArtifactByUrlAndVersion` function for getting child artifacts from relatedArtifacts, added `batchDraft` for creating draft artifacts in a batch transaction
- `service/src/requestSchemas.ts` - additional schemas for enforcing `$draft` parameters (must have `id` and `version` and `version` must be in the format `MAJOR.MINOR.PATCH.REVISION`
- `service/src/services/LibraryService.ts` / `service/src/services/MeasureService.ts` - add `draft` endpoint
- `service/src/types/service-types.ts` - new types `CRMILibrary` and `CRMIMeasure`
- `service/src/util/bundleUtils.ts` - `createBatchResponseBundle` function for `$draft` operation output 
- `service/src/util/inputUtils.ts` - adds the following helper functions to validate inputs: `checkDraft`, `checkIsOwned`, and `checkExistingArtifact`
- `service/src/util/serviceUtils.ts` - adds the following helper functions for `$draft`: `getChildren` and `modifyResourcesForDraft`
- `LibraryService.test.ts` / `MeasureService.test.ts` / `serviceUtils.test.ts` - added unit tests

# Testing guidance
- `npm run check:all`
- `npm run build:all`
- `cd service`
- `npm run db:reset`
- `npm run db:loadBundle <your favorite bundles>` (I also recommend using the double nested child artifact bundle I "made" so that the recursion can be properly tested.
- `npm run start:all`
- The attached zip file contains Insomnia requests for testing!

[DraftTesting.json](https://github.com/user-attachments/files/16089632/DraftTesting.json)

### DIFF
--- a/service/src/requestSchemas.ts
+++ b/service/src/requestSchemas.ts
@@ -3,6 +3,7 @@ import { z } from 'zod';
 import { BadRequestError, NotImplementedError } from './util/errorUtils';
 
 const DATE_REGEX = /([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1]))?)/;
+const VERSION_REGEX = /^\d+\.\d+\.\d+\.\d+$/;
 
 // Operation Definition: http://hl7.org/fhir/us/cqfmeasures/STU4/OperationDefinition-cqfm-package.html
 const UNSUPPORTED_PACKAGE_ARGS = [
@@ -52,6 +53,8 @@ const UNSUPPORTED_LIBRARY_SEARCH_ARGS = [...UNSUPPORTED_CORE_SEARCH_ARGS, 'conte
 
 const hasIdentifyingInfo = (args: Record<string, any>) => args.id || args.url || args.identifier;
 
+const idAndVersion = (args: Record<string, any>) => args.id && args.version;
+
 /**
  * Returns a function that checks if any unsupported params are present, then runs the
  * other passed in functions in sequence. Each catchFunction is expected to check for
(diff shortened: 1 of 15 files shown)


service/src/requestSchemas.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 import { BadRequestError, NotImplementedError } from './util/errorUtils';
 
 const DATE_REGEX = /([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1]))?)/;
+const VERSION_REGEX = /^\d+\.\d+\.\d+\.\d+$/;
```
hossenlopp, line 6: This regex makes it require the full semantic version including the 4th "revision" spot. It is implied that is required in $draft definition but elsewhere in CRMI they use only 3 parts to the "patch" spot. This should allow for the revision spot to be optional.

```suggestion
const VERSION_REGEX = /^\d+\.\d+\.\d+(\.\d+|)$/;
```

hossenlopp (changes requested):
Initial review.  Good reuse of a lot of logic from the frontend. The version regex check seems to be too specific.

Will continue reviewing on Monday.

hossenlopp (approved):
Looks good and is good at checking invalid input and version conflicts. Nice re-work of code from the app for this.

PR #104, opened 2024-07-27
Expose clone operation

elsaperelli:
# Summary
This PR exposes the $clone operation as an endpoint on the server. This functionality is defined in the CRMI IG [here](https://build.fhir.org/ig/HL7/crmi-ig/OperationDefinition-crmi-clone.html).

## Note
The current OperationDefinition of the clone operation does not include `url` as an input parameter. This is an error and is being tracked in [this](https://jira.hl7.org/browse/FHIR-46385) Jira issue by Paul. Additionally, the spec needs to detail how child artifacts will be handled for clone. For now, I made the decision to simply add `-clone` to the end of the child artifact url when its parent is cloned. This approach made the most sense to me since there is no current guidance for it and the url cannot stay the same but I am open to suggestions!

## New behavior
The user can now clone a Measure or Library artifact and any resources it is composed of regardless of status by sending a GET or POST request to `Measure/$clone`, `Measure/:id/$clone`, `Library/$clone` and `Library/:id/$clone`. Since this is only an operation supported in an Authoring Artifact Repository, the `AUTHORING` environment variable must be set to true.

## Code changes
- `service/src/config/serverConfig.ts` - added `$clone` endpoints to server configuration
- `service/src/config/capabilityStatementResources.json` - adds draft and clone operations to capability statement
- `service/src/db/dbOperations.ts` - added `batchClone` for cloning artifacts in a batch transaction
- `service/src/requestSchemas.ts` - additional schemas for enforcing `$clone` parameters (must have `id`, 'version', and `url`)
- `service/src/services/LibraryService.ts` / `service/src/services/MeasureService.ts` - add `clone` endpoint
- `service/src/util/inputUtils.ts` - change name of function 
- `service/src/util/serviceUtils.ts` - adds helper function `modifyResourcesForClone`
- `LibraryService.test.ts` / `MeasureService.test.ts` - added unit tests

# Testing guidance
- `npm run check:all`
- `npm run build:all`
- `cd service`
- `npm run db:reset`
- `npm run db:loadBundle <your favorite bundle>` (I also recommend using the double nested child artifact bundle I "made" so that the recursion can be properly tested.
- `npm run start:all`
- The attached zip file contains Insomnia requests for testing! (assuming database is loaded with the nested child artifact bundle)

[Clone-Testing.json](https://github.com/user-attachments/files/16418565/Clone-Testing.json)

### DIFF
--- a/service/src/util/serviceUtils.ts
+++ b/service/src/util/serviceUtils.ts
@@ -92,4 +92,33 @@ export async function modifyResourcesForDraft(artifacts: FhirArtifact[], version
  * Helper function that takes an active or draft artifact and returns it with a new id,
  * same status, url (from the $clone parameter), and version (from the $clone parameter).
  */
-export async function modifyResourcesForClone(artifacts: FhirArtifact[], url: string, version: string) {}
+export async function modifyResourcesForClone(artifacts: FhirArtifact[], url: string, version: string) {
+  for (const artifact of artifacts) {
+    artifact.id = uuidv4();
+    artifact.version = version;
+    artifact.url = url;
+    if (artifact.relatedArtifact) {
+      artifact.relatedArtifact.forEach(ra => {
+        if (
+          ra.type === 'composed-of' &&
+          ra.resource &&
+          ra.extension?.some(
+            e => e.url === 'http://hl7.org/fhir/StructureDefinition/artifact-isOwned' && e.valueBoolean === true
+          )
+        ) {
+          const oldUrl = ra.resource.split('|')[0];
+          ra.resource = oldUrl + '|' + version;
+        }
+      });
+    }
+
+    const checkExisting = await findArtifactByUrlAndVersion(artifact.url, version, artifact.resourceType);
+    if (checkExisting) {
+      throw new BadRequestError(
+        `A ${artifact.resourceType} with url ${artifact.url} and version ${artifact.version} already exists in the database.`
+      );
+    }
+  }
+
+  return artifacts;
+}
--- a/service/src/util/serviceUtils.ts
+++ b/service/src/util/serviceUtils.ts
@@ -92,11 +92,10 @@ export async function modifyResourcesForDraft(artifacts: FhirArtifact[], version
  * Helper function that takes an active or draft artifact and returns it with a new id,
  * same status, url (from the $clone parameter), and version (from the $clone parameter).
  */
-export async function modifyResourcesForClone(artifacts: FhirArtifact[], url: string, version: string) {
+export async function modifyResourcesForClone(artifacts: FhirArtifact[], version: string) {
   for (const artifact of artifacts) {
     artifact.id = uuidv4();
     artifact.version = version;
-    artifact.url = url;
     if (artifact.relatedArtifact) {
       artifact.relatedArtifact.forEach(ra => {
         if (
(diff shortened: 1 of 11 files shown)


service/src/util/serviceUtils.ts
```diff
@@ -87,3 +87,37 @@ export async function modifyResourcesForDraft(artifacts: FhirArtifact[], version
...
+            e => e.url === 'http://hl7.org/fhir/StructureDefinition/artifact-isOwned' && e.valueBoolean === true
+          )
+        ) {
+          const oldUrl = ra.resource.split('|')[0];
+          ra.resource = oldUrl + '-clone|' + version;
+        }
+      });
+    }
+
+    const checkExisting = await findArtifactByUrlAndVersion(artifact.url, version, artifact.resourceType);
```
lmd59, line 114: Doesn't the database already check uniqueness by url and version when inserting a new artifact?

lmd59 (commented):
Still need to do some testing now that my environment is fixed, but here are some initial comments on code!
Also might be worth adding something to the service folder's README file!

service/src/util/serviceUtils.ts
elsaperelli: I think it only does by id? Happy to discuss more

lmd59: https://github.com/projecttacoma/measure-repository/blob/f876096accea79defc79c03a8a0c8842421c51e2/service/scripts/dbSetup.ts#L21 has a url/version index as well.

lmd59: Did a quick look at removing the checkExisting bits, and the server appropriately sends the duplicate key error: 
"Unexpected: E11000 duplicate key error collection: measure-repository.Measure index: url_1_version_1 dup key: { url: \"http://example1.com\", version: \"1.0.0\" }"
If we want to do a Bad Request error instead, it might be cleaner to catch the database error and rethrow a Bad Request from it.

lmd59 (commented):
Updates look good! Happy to talk through the url/version duplication when you're back.

service/src/util/serviceUtils.ts
lmd59: @hossenlopp - maybe you have thoughts on this?

hossenlopp: Yeah. This error should be caught and turned into a bad request. If we can identify the error we could give feedback that this resource already exists.

elsaperelli: The `findArtifactByUrlAndVersion` function throws a BadRequest error to give feedback that this resource already exists and this happens before the batch insertion happens. Do you want me to move this to `batchClone` and `batchDraft` (where those database insertions are occurring)? Those insertions are already in a try catch so should I check that it is a duplicate key error specifically?

lmd59: I'm thinking that a duplicate key error would typically be a bad request across the board (in all cases, not just this one, but interested if other folks agree), so maybe we can strip out the checks here, leave it as a straight duplicate key for now, and create a new task for turning duplicate key into bad request error across the board.

elsaperelli: Yeah I think that's a good idea considering we currently don't wrap any duplicate key errors in a BadRequestError.

lmd59 (approved):

PR #106, opened 2024-08-15
Expose  operation

elsaperelli:
# Summary
This PR exposes the $approve operation as an endpoint on the server. This functionality is defined in the CRMI IG [here](https://hl7.org/fhir/uv/crmi/OperationDefinition-crmi-approve.html).

## New behavior
The user can now approve a Measure or Library artifact and any resources it is composed of regardless of status by sending a GET or POST request to `Measure/$approve`, `Measure/:id/$approve`, `Library/$approve` and `Library/:id/$approve`. This operation takes three optional parameters: `approvalDate`, `artifactAssessmentType`, and `artifactAssessmentSummary`. If `artifactAssessmentType` and `artifactAssessmentSummary` are both provided, then a [cqf-artifactComment](https://hl7.org/fhir/extensions/5.1.0/StructureDefinition-cqf-artifactComment.html) extension will be added to the artifact and any of its children. If `approvalDate` is provided, the `approvalDate` is set, if not, it is set to the system. The `date` is also set to the system date. Since this is only an operation supported in an Authoring Artifact Repository, the `AUTHORING` environment variable must be set to true.

## Code changes
- `service/README.md` - add description of `$approve`
- `service/src/config/capabilityStatementResources.json` - add crmi-approve
- `service/src/config/serverConfig.ts` - add `approve` endpoints
- `service/src/db/dbOperations.ts` - add `batchUpdate` function
- `service/src/requestSchemas` - add id and type and summary checks and ApproveArgs for type checking
- `service/src/services/LibraryService.ts` / `service/src/services/MeasureService.ts` - add `approve` function
- `service/src/util/inputUtils.ts` - add `valueCode` to `gatherParams`
- `service/test/services/LibraryService.test.ts` / `service/test/services/MeasureService.test.ts` - added unit tests 

# Testing guidance
- `npm run build:all`
- `npm run check:all`
- In `service`, reset the database (`npm run db:reset`) and test with the nested child artifact bundle I made and may connectathon bundles (attached example Insomnia requests are for use with the created nested artifact bundle
- `npm run start:all`
- Go through example Insomnia requests. Make sure `date` and `approvalDate` are changed/added, `extension` is added if `artifactAssessmentSummary` AND `artifactAssessmentType` are included as input parameters, etc.

[approve-testing.json](https://github.com/user-attachments/files/16626948/approve-testing.json)

PR #107, opened 2024-08-21
Expose review

lmd59:
# Summary
Implements the `$review` endpoint in the server according to the [CRMI IG definition](https://hl7.org/fhir/uv/crmi/OperationDefinition-crmi-review.html).

## New behavior

For all details, this functionality is essentially the same as `$approve` with the exception of the `approvalDate` parameter, which becomes the `reviewDate`, and instead of setting the `approvalDate` field on the artifact, it sets the `lastReviewDate` field.

## Code changes

See https://github.com/projecttacoma/measure-repository/pull/106

# Testing guidance

Use the testing guidance in https://github.com/projecttacoma/measure-repository/pull/106, including Insomnia requests, but change all `$approve` to `$review` and all `approvalDate` parameters to `reviewDate`.

### DIFF
--- a/service/src/services/LibraryService.ts
+++ b/service/src/services/LibraryService.ts
@@ -337,7 +337,7 @@ export class LibraryService implements Service<CRMIShareableLibrary> {
     return createBatchResponseBundle(approvedArtifacts);
   }
 
-    /**
+  /**
    * result of sending a POST or GET request to:
    * {BASE_URL}/4_0_1/Library/$review or {BASE_URL}/4_0_1/Library/[id]/$review
    * applies a review to an existing artifact, regardless of status, and sets the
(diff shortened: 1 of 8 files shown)


service/src/services/LibraryService.ts
```diff
@@ -336,6 +337,82 @@ export class LibraryService implements Service<CRMIShareableLibrary> {
     return createBatchResponseBundle(approvedArtifacts);
   }
 
+  /**
+   * result of sending a POST or GET request to:
+   * {BASE_URL}/4_0_1/Library/$review or {BASE_URL}/4_0_1/Library/[id]/$review
+   * applies a review to an existing artifact, regardless of status, and sets the
+   * date and lastReviewedDate elements of the reviewed artifact as well as for all resources
```
elsaperelli, line 344:
```suggestion
   * date and lastReviewDate elements of the reviewed artifact as well as for all resources
```

service/src/services/LibraryService.ts
```diff
@@ -336,6 +337,82 @@ export class LibraryService implements Service<CRMIShareableLibrary> {
...
+        library.extension.push(comment);
+      } else {
+        library.extension = [comment];
+      }
+    }
+    library.date = new Date().toISOString();
+    library.lastReviewDate = parsedParams.reviewDate ?? new Date().toISOString();
+    checkIsOwned(library, 'Child artifacts cannot be directly reviewed.');
+
+    // recursively get any child artifacts form the artifact if they exist
```
elsaperelli, line 388:
```suggestion
    // recursively get any child artifacts from the artifact if they exist
```

elsaperelli: This is also my typo from the $approve function if you don't mind changing that too :)

service/src/services/MeasureService.ts
```diff
@@ -338,6 +339,82 @@ export class MeasureService implements Service<CRMIShareableMeasure> {
     return createBatchResponseBundle(approvedArtifacts);
   }
 
+  /**
+   * result of sending a POST or GET request to:
+   * {BASE_URL}/4_0_1/Measure/$review or {BASE_URL}/4_0_1/Measure/[id]/$review
+   * applies a review to an existing artifact, regardless of status, and sets the
+   * date and lastReviewedDate elements of the reviewed artifact as well as for all resources
```
elsaperelli, line 346:
```suggestion
   * date and lastReviewDate elements of the reviewed artifact as well as for all resources
```

elsaperelli (changes requested):
Just some small comments but looks great!

service/src/services/LibraryService.ts
lmd59: Will do!

PR #108, opened 2024-08-27
Handle duplicate key errors from mongo and return Bad Request errors

hossenlopp:
# Summary

Currently when a database change is made that would cause a duplicate `url` and `id` to occur, the response from the server is 500 from mongo enforcing the restraint. This work catches the specific mongo errors responds with a 400 Bad Request `invalid` response explaining the issue. 

## New behavior

All DB operations that make changes will catch the duplicate key errors from mongo and throw a BadRequestError that will result in a 400 response.

Adds collection index creation to the test db setup functions and fixes many async issues with the unit tests.

## Code changes

`dbOperations.ts` - Added a helper function to convert a caught error to a BadRequestError if it happens to be the mongo duplicate key error.
`tests/utils.ts` - Added index creation to `setupTestDatabase`.
Unit Tests - Fixes to many of the test database setup and tear down to ensure the database is actually ready before tests. Added to test to operations that change/insert resources to check for 400 error on duplicate key situations.

# Testing guidance

Run unit tests.

Attempt to insert a resource that has the same url and version as one that is already in your database.

### DIFF
--- a/service/src/db/dbOperations.ts
+++ b/service/src/db/dbOperations.ts
@@ -97,15 +97,7 @@ export async function createResource(data: FhirArtifact, resourceType: string) {
     await collection.insertOne(data);
     return { id: data.id as string };
   } catch (e) {
-    if (e instanceof MongoServerError && e.code === 11000) {
-      let errorString = 'Resource with primary identifiers already in repository.';
-      if (e.keyPattern) {
-        errorString =
-          'Resource with identifiers (' + Object.keys(e.keyPattern).join(',') + ') already exists in the repository.';
-      }
-      throw new BadRequestError(errorString);
-    }
-    throw e;
+    throw handlePossibleDuplicateKeyError(e);
   }
 }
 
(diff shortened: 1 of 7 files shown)


service/src/db/dbOperations.ts
```diff
@@ -174,10 +182,22 @@ export async function batchUpdate(artifacts: FhirArtifact[], action: string) {
...
   if (error) throw error;
   return results;
 }
+
+function handlePossibleDuplicateKeyError(e: any) {
+  if (e instanceof MongoServerError && e.code === 11000) {
+    let errorString = 'Resource with primary identifiers already in repository.';
+    if (e.keyPattern) {
+      errorString =
+        'Resource with identifiers (' + Object.keys(e.keyPattern).join(',') + ') already exists in the repository.';
```
lmd59, line 198: Should the error message identify what url/version are duplicating? This might help especially when multiple things are being inserted/updated in succession.
For instance, the dup key error looks something like this: 
"Unexpected: E11000 duplicate key error collection: measure-repository.Measure index: url_1_version_1 dup key: { url: \"SevereObstetricComplicationsFHIR-test\", version: \"SevereObstetricComplicationsFHIR\" }"
We could even stick the mongo error message in as part of our new error string.

lmd59 (commented):
Looks good, and thank you for making these updates so that our tests are more consistent!
One small comment!

service/src/db/dbOperations.ts
hossenlopp: Added the key info and the resource type to the error message.

PR #116, opened 2024-11-25
Change service location in frontend, another upload script

elsaperelli:
# Summary
This PR changes the service location URL in the frontend app to no longer have the `/metadata` endpoint. It also adds a new db option: `loadPublishableContent`. This essentially does the same thing as `loadBundle` (takes a measure bundle or directory of measure bundles and uploads them to the measure repository service); however, the one difference is that if there are any `draft` artifacts in the measure, they will be coerced to `active` so they can be uploaded without error.

## New behavior
- `npm run db:loadPublishableContent` script

## Code changes
- `README.md` - additions regarding new db upload option
- `package.json` - `npm run db:loadPublishableContent`
- `app/src/pages/index.tsx` - remove `/metadata`
- `service/scripts/dbSetup.ts` - `loadPublishableContent` now allows measure bundle and measure bundle directory upload while coercing artifacts to `active`

# Testing guidance
- `cd service`
- `npm run db:reset`
- `npm run db:loadContentBundle <path to one of the bundles where Measure has status draft>`
- `cd ..`
- `npm run start:all`
- Make sure Measure has been uploaded with status `active`
- Check Service Location URL in frontend

### DIFF
--- a/app/src/pages/index.tsx
+++ b/app/src/pages/index.tsx
@@ -67,7 +67,7 @@ export default function Home({
       <Divider my="sm" variant="dotted" />
       <Title order={2}>Service Location:</Title>
       <div style={{ marginTop: '18px', marginBottom: '18px' }}>
-        <Anchor href={`${serviceUri}/metadata`}>{`${serviceUri}/metadata`}</Anchor>
+        <Anchor href={`${serviceUri}`}>{`${serviceUri}`}</Anchor>
       </div>
       <Divider my="sm" variant="dotted" />
       <Title order={2}>Service Capabilities:</Title>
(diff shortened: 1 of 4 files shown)


lmd59 (commented):
What are your thoughts to having this script load the environment variable and do this publishable coercion based on that rather than the script name?

elsaperelli:
> What are your thoughts to having this script load the environment variable and do this publishable coercion based on that rather than the script name?

Hmm interesting! At least I was going to say yes, but my only hesitation is the following scenario: technically these MADiE bundles should have measures with status `active` so even if we do allow authoring, there are cases where we still want to upload all the measures and libraries in a MADiE bundle to the publishable side and be active. Does that make sense? Happy to discuss more.

PR #121, opened 2025-01-15
Cleaned up docker build shell script and warnings in Dockerfiles

hossenlopp:
# Summary

## Code changes

# Testing guidance

### DIFF
--- a/app.Dockerfile
+++ b/app.Dockerfile
@@ -1,6 +1,6 @@
-FROM node:18 as base
+FROM node:18 AS base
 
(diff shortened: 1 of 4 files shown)


elsaperelli (changes requested):
I still get a warning about `version` being obsolete, what warnings does this PR specifically fix so I can test?

elsaperelli (changes requested):
I would just also make these changes in the `docker-compose.example.yml` file!

PR #122, opened 2026-04-07
Update to Node 24 and Use Hardened Docker Images

hossenlopp:
# Summary
Updates CI to use node 24. Update docker builds to make use of dhi.io based images. npm audit fixes.

## New behavior
None.

## Code changes
- docker-compose.yml - No longer referencing docker hub images, sorted out replica set issue
- Dockerfiles - Using hardened node images

# Testing guidance
- `npm run check:all`
- `docker compose up --build` - Ensure measures can be pushed to the measure repository and accessed in the app.


service.Dockerfile
```diff
@@ -33,19 +32,16 @@ RUN rm -rf node_modules
...
 
 COPY --from=build --chown=node:node /home/node/app/service/dist* ./dist
 COPY --from=build --chown=node:node /home/node/app/node_modules ./node_modules
 
-
 # Start app
 EXPOSE 3000
-ENV PORT 3000
-ENV HOST "0.0.0.0"
+ENV PORT=3000
```
lmd59, line 45:
```suggestion
ENV AUTHORING="true"
ENV PORT=3000
```
Maybe this is a little out of scope for the current PR, but something I noticed is that we should probably have an environment variable value for `AUTHORING` since I worry we might have some inconsistent behavior if this environment variable isn't set at all. My recommendation for a default value is true - mostly because it's the more interesting set of functionality.

lmd59 (commented):
Things looks good! I had one recommendation that I think we should include in this PR if you don't have any concerns.

service.Dockerfile
hossenlopp: Sounds good! changed in [8cb23f8](https://github.com/projecttacoma/measure-repository/pull/122/commits/8cb23f8d6235592688e2d25d87cdd36dd1b716a8)

lmd59 (approved):

